PR #13, opened 2014-06-06
[output] Add a minimal output option for the middleware

alexhancock:
This adds another option for output to the middleware giving you the....

```
webpack: bundle is now VALID/INVALID
```

... messages without the full output of the compilation stats.

It's a middle ground between the default and the `noInfo` and `quiet` options.

I wonder if a better general solution would be to add the concept of a `logLevel` option, and have something like...

```
logLevel: "QUIET" // -> Currently quiet
logLevel: "WARNINGS" // -> Currently noInfo
logLevel: "MINIMAL" // -> Currently minimalOutput
logLevel: "DEBUG" // -> Currently the default
```

I can investigate that if you're interested.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -10,7 +10,7 @@ var mime = require("mime");
 module.exports = function(compiler, options) {
 	if(!options) options = {};
 	if(options.watchDelay === undefined) options.watchDelay = 200;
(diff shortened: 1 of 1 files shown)


lili21:
where is the doc about stats option ?

PR #22, opened 2015-01-30
Added "watch" config flag to allow middleware to only compile once

danrspencer:
Also updated docs to include the "filename" option which is used during lazy mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,12 +47,20 @@ app.use(webpackMiddleware(webpack({
 	// switch into lazy mode
 	// that means no watching, but recompilation on every request
 
+	watch: true,
+	// disable to turn off watching and only compile
+	// when the server is started (only lazy: false)
+
 	watchDelay: 300,
-	// delay after change (only lazy: false)
+	// delay after change (only lazy: false && watch: true)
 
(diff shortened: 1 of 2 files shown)


sokra:
There is no watch option. `lazy` means no `watch`.

Could you make it more clear that `filename` is **only** used in lazy mode?

danrspencer:
Adding the watch boolean is to allow both watch and lazy modes to be disabled. E.g. only compile once

I can update the filename docs if necessary but I was just attempting to update the docs to reflect existing functionality, I haven't changed anything around that myself (found on during testing that the undocumented filename option was required otherwise lazy mode didn't work).

sokra:
Why would you compile only once? This looks like an production usecase.

You are correct about the `filename` documentation.

danrspencer:
It is a "production" usecase, however in this case the app is an internal tool where we're with the risks of the dev middlewhere. Flipping the "watch" switch allows us to change between dev and live easily without having to have slightly different code paths between the two.

SpaceK33z:
What @sokra said, this is a production usecase. This middleware is not meant for that.

PR #26, opened 2015-02-11
add Content-Length header to response

dtinth:
When bundle is large, I download the script using XMLHttpRequest, so
that the download progress can be monitored.

During development, to test that this works requires that Content-Length
header be sent in the dev server.


sokra:
Thanks

PR #31, opened 2015-05-11
support build into filesystem

ssnau:
Hi,

I notice the middleware will replace the outputFileSystem with an in-memory one. But sometimes i may want to build them into my static  directory so that i can write relative path to request the script while still benefit features (such as HMR) from the dev server.

```
<script src="/path/to/script.js"></script>
```

should we just add an option named: `useRealFileSystem`?

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -15,7 +15,7 @@ module.exports = function(compiler, options) {
 
 	// store our files in memory
 	var files = {};
-	var fs = compiler.outputFileSystem = new MemoryFileSystem();
+	var fs = compiler.outputFileSystem = (options.useRealFileSystem && compiler.outputFileSystem) || new MemoryFileSystem();
 
 	compiler.plugin("done", function(stats) {
 		// We are now on valid state


sokra:
no, this is not good. You shouldn't access the file on the disk.

There are other options that allow to write a relative path, i. e. `proxy` if you depend on this.

jbrantly:
> There are other options that allow to write a relative path

This is only true if you are hosting the HTML file from the middleware. If you're trying to combine with an existing server instead you currently have to do this ([from the documentation](http://webpack.github.io/docs/webpack-dev-server.html#combining-with-an-existing-server)):

> On the other side you need to teach your backend server to generate HTML pages that include script tags that point to assets on the webpack-dev-server. 

If the middleware also wrote to disk then that would no longer be an absolute requirement as the backend server could serve the relative path from disk like usual for the initial load but then hot replacements could still be served by the middleware. This would greatly simplify generating the HTML between production and development.

I would ask that you reconsider this feature.

use-strict:
@sokra , I also need this for a different use case. This memory file-system forces me to serve the bundles from the dev-server and in my case this simply isn't possible due to project setup. I would really like not having to use forked versions of the dev-server/middleware for this. I also need the physical files during development for other purposes.

I see no reason why this should be restricted. Please reconsider merging this PR.

koistya:
@sokra I do also need this feature for [react-starter-kit](https://github.com/kriasoft/react-starter-kit), I think it will unclock a couple of good development workflows.

krazik:
+1 also need this.

PR #32, opened 2015-05-15
Use watchOptions instead of solely allowing watchDelay

janraasch:
Use case:

```js
watchOptions: {
  poll: true
}
```

in `webpack.config.js`.

This is analogous to the deprecation in the `webpack` CLI.

See https://github.com/webpack/webpack/blob/89058a2c4a32a2ec31c88acb3789c4f9eda58db1/bin/convert-argv.js#L84 and https://github.com/webpack/webpack/blob/89058a2c4a32a2ec31c88acb3789c4f9eda58db1/lib/webpack.js#L35

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -9,7 +9,13 @@ var mime = require("mime");
 // constructor for the middleware
 module.exports = function(compiler, options) {
 	if(!options) options = {};
-	if(options.watchDelay === undefined) options.watchDelay = 200;
+	if(typeof options.watchOptions === "undefined") options.watchOptions = {};
+	if(typeof options.watchDelay !== "undefined") {
+		// TODO remove this in next major version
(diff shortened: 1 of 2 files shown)


middleware.js
```diff
@@ -66,7 +72,7 @@ module.exports = function(compiler, options) {
 	var state = false;
 
 	// in lazy mode, rebuild automatically
-	var forceRebuild = false;
+	var forceRebuild = false;de
```
sokra: de?

janraasch: whoops.

janraasch: @sokra Fixed. Sorry about that...

maxcnunes:
Any plan merging it to master? I would really appreciate use the `watchOptions`.

sokra:
Thanks

maxcnunes:
Thanks :smile:

PR #36, opened 2015-07-01
Add progress support

shepherdwind:
Hi, this package is very useful when develop with koa or expressjs.

But, every time server restart, there is a long time console do not output anything. When I use the grunt-webpack, there is progress message, that make me feel more speed.

So, I think this package need progress too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,9 @@ app.use(webpackMiddleware(webpack({
 	quiet: false,
 	// display nothing to the console
 
+	progress: false,
+  // show bundle build progress
+
 	lazy: true,
 	// switch into lazy mode
 	// that means no watching, but recompilation on every request
--- a/middleware.js
+++ b/middleware.js
@@ -5,6 +5,7 @@
 var path = require("path");
 var MemoryFileSystem = require("memory-fs");
 var mime = require("mime");
+var ProgressPlugin = require("webpack/lib/ProgressPlugin");
 
 // constructor for the middleware
 module.exports = function(compiler, options) {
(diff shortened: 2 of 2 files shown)


romseguy:
Why was this closed? Looks useful to me

gpbl:
Yeah why no progress?

rublev:
+1 - would like compile progress.

shepherdwind:
The project seems no more maintenance .

There is an example below, which have the same effect on my merge request.

```js
const join = require('path').join;
const assign = require('object-assign');
const ProgressPlugin = require('webpack/lib/ProgressPlugin');
const webpack = require('webpack');
const fs = require('fs');
const webpackMiddleware = require('koa-webpack-dev-middleware');
const getPackageName = require('./getPackageName');

module.exports = function() {
  const appName = getPackageName();

  let webpackConfig;
  const webpackConfigFile = join(process.cwd(), 'config/webpack.dev.config.js');
  if (fs.existsSync(webpackConfigFile)) {
    webpackConfig = require(webpackConfigFile);
  } else {
    webpackConfig = require('./webpack.dev.config.js');
  }

  // noInfo only show error message
  const cfg = assign({}, {
    publicPath: '/' + appName + '/',
    noInfo: true,
    progress: true,
  });

  if (cfg.progress) {
    compiler.apply(new ProgressPlugin(function(percentage, msg) {
      const stream = process.stderr;
      if (stream.isTTY && percentage < 0.71) {
        stream.cursorTo(0);
        stream.write(msg);
        stream.clearLine(1);
      } else if (percentage === 1) {
        console.log('\n[chair-webpack]: bundle build is now finished.');
      }
    }));
  }

  return webpackMiddleware(compiler, cfg);
};
```

tnguyen14:
@shepherdwind Can this PR be reopened? The project seems to be active still now?

shellscape:
happy to take a look at it if it is

PR #41, opened 2015-10-06
Include output path in file name pattern

jugglinmike:
When the `lazy` option is set, the `filename` pattern is used to
determine which requests should trigger a reload. Because the string
being tested in that case includes the output path (see the
`getFilenameFromUrl` function), the matching pattern should also include
it.

Update the output path in the `filename` pattern so that the `lazy`
option correctly enables building on request.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -23,7 +23,7 @@ module.exports = function(compiler, options) {
 			var str = options.filename
 				.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")
 				.replace(/\\\[[a-z]+\\\]/ig, ".+");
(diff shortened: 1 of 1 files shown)


bebraw:
:+1:

SpaceK33z:
@jugglinmike, interested in fixing the merge conflicts? I'll test it and merge after that.

SpaceK33z:
@jugglinmike ping

SpaceK33z:
I'm closing this since nobody seems to have this issue, the author is not active anymore, and it has a couple of problems:

- It causes a breaking change for people who have defined an output path.
- `compiler.outputPath` is added in the regex constructor, but it should be added in the `str` variable. Otherwise it will not be properly escaped.
- Needs a test.

PR #44, opened 2015-11-17
Add docs around using History API

### LINKED ISSUE #39: Support History API urls
Hey there,

I am using `webpack-dev-middleware` to develop a SPA, with an `index.html` which is programatically generated by webpack. With this plugin, if I navigate to anything other than `/` or `/index.html`, the middleware does not pick up the request, and the request fails.

For most people, something like this solves the problem:

```js
app.get('*', function (req, res) {
  res.send(path.join(__dirname, 'index.html'))
}
```

but for previously mentioned reasons, this doesn't work. I have written this handler to catch urls missed by `webpack-dev-middleware`, but would look at adding this into the core.

```js
app.use(function (req, res, next) {
  if (req.url.match(/^(\/[a-zA-Z-]+)+(.html)?(\?.+)?$/)) { // Just a regex I spun up to sort my use case
    devMiddleware(
      {
        url: '/index.html'
      },
      res,
      next
    )
  } else {
    next()
  }
})
```

Forgive me if I've completely missed a config flag or something - I have looked at the source and couldn't find anything.

I am happy to help write a PR to support this, especially since this will become more popular with `react-transform` becoming more popular.

Thanks for your help in advance

frederickfogerty:
Closes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,4 +79,4 @@ const middleware = require('webpack-dev-middleware')(compiler)
 app.get('*', (req, res) => res.end(middleware.fileSystem.readFileSync(path.join(config.output.path, 'index.html')))
 ```
 
-See more in this [gist](https://gist.github.com/frederickfogerty/df921f21a83b479b2056). For a more simple version, see [here](https://gist.github.com/maman/8e91e6cb4ca9feaa4290). Also see [issue #39](https://github.com/webpack/webpack-dev-middleware/issues/39) for discussion.
+See for some examples check out the [html5-history folder](//github.com/webpack/webpack-dev-middleware/examples/html5-history/) in examples. Also see [issue #39](//github.com/webpack/webpack-dev-middleware/issues/39) for discussion.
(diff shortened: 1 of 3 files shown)


bebraw:
Can you push those Gist examples below a `examples/` directory? The problem with gists is that they might disappear.

Apart from that this looks good to me.

frederickfogerty:
@bebraw done! :tada:

bebraw:
@sokra Safe to merge?

sokra:
Nope, not safe. It's the wrong way. It's unsafe to read from the webpack filesystem directly. The webpack-dev-middleware is build around to concept that requests block until compilation has finished. Reading directly would bypass this concept.

Instead of requesting from filesystem you should better rewrite the path of the request and give the rewritten request to the webpack-dev-middleware (which handles the blocking). Or just use the `connect-history-api-fallback`, which does exactly that. Just put the webpack-dev-middleware before and after this middleware. (Before to handle the raw request and after to handle the rewritten request)

frederickfogerty:
Funnily enough, that methodology what was I used to before I adopted this
method!

nickretallack:
@sokra can you elaborate on that?  We're supposed to add the webpackDevMiddleware twice?  Should the two instances of it differ?

PR #49, opened 2015-11-27
Allow to customize `outputPath`

koistya:
Currently there is no way to make `webpack-dev-middleware` work with multiple compilers, where each compiler has its own `outputPath`.

For example, server-side bundle is written to `/build/server.js` and client-side bundle is written to `/build/public/client.js`. The current version of middleware tries to find `client.js` in the `/build` folder. Note that, Node.js app (`server.js`) serves static files from the `public` folder, e.g. there are links such as `<script src="/client.js">` and not `<script src="/public/client.js">`.

##### Usage sample:

```js
webpackDevMiddleware(compiler, {
  outputPath: path.join(__dirname, '../build/public');
});

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ app.use(webpackMiddleware(webpack({
 	// public path to bind the middleware to
 	// use the same as in webpack
 
+	outputPath: __dirname + "/build/public",
+	// output path, in case it differs from the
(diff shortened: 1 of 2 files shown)


koistya:
Or, even better consider merging PR https://github.com/webpack/webpack-dev-middleware/pull/50 instead.

PR #50, opened 2015-11-27
Fix setting in-memory file stream and base output path

koistya:
Set in-memory file stream and base output path only for `web` and `webwoker` compilers.

This way you can make it work with `MultiCompiler`, containing compilers with both `web` and `node` targets. It will correctly resolve files such as `<script "/client.js">` located in `/build/public/client.js` folder. Also it will preserve server-side bundle (`/build/server.js`) on disk, so that you could launch Node.js server from it.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -29,7 +29,32 @@ module.exports = function(compiler, options) {
 
 	// store our files in memory
 	var files = {};
-	var fs = compiler.outputFileSystem = new MemoryFileSystem();
+	var fs = new MemoryFileSystem();
+
+	// the base output path for web and webworker bundles
+	var outputPath;
+
+	if (typeof compiler.compilers === "undefined") {
+		compiler.outputFileSystem = fs;
+		outputPath = compiler.outputPath;
+	} else {
+		// when there is a mix of web and non-web compilers
+		// set the base outputPath and in-memory file stream
+		// only for the web and webwoker compilers
+		outputPath = compiler.compilers[0].outputPath;
+		compiler.compilers[0].outputFileSystem = fs;
+		for(var i = 1; i < compiler.compilers.length; i++) {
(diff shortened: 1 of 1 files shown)


sokra:
This relies on too many internal details: `.compilers` and the name of the targets are internal details that should not be knows by the middleware.

It makes no sense to use target: "node" with the middleware. For this case you better start two compilers: one with the middleware and target web, and one with target node.

It's not a valid use case to use the middleware with target node.

koistya:
@sokra in the context of an isomorphic web app, the majority of the code is going to be reused. Do you suggest **(1)** to initialize two compilers, one for Node.js bundle and another for client-side bundle(s)? Or, **(2)** create a single compiler (`MultiCompiler`) and start inner compilers separately (the ones with target `web` via `webpack-dev-middleware`, and the one with target `node` by something like this `bundler.compilers.filter(x => x.options.target === 'node').forEach(x => x.watch(..))`.

sokra:
> (1) to initialize two separate compilers, one for Node.js bundle and another for client-side bundle(s)? This will significantly reduce performance when this two bundles need to work simultaneously?

You can set the `inputFileSystem` on both compilers to the same instance.

``` js
compiler2.inputFileSystem = compiler1.inputFileSystem;
```

This way they will share disk cache. Watcher are statically shared (they only need to be in the same node.js process).

chrisblossom:
@sokra I have tried your suggestion by setting the inputFileSystems the equal but have run into an issue. If compiler2 has a file changed that is outside of compiler1's tree, the change is not written to disk unless it is significant in size.

For example, given the entry points client.js and server.js...
When there is a minor change to server.js webpack will not write the updated server.js file. Works as expected with files that both client.js and server.js have in their tree.

Any ideas?

PR #55, opened 2015-12-17
Fix serving index.html on Windows

### LINKED ISSUE #54: On Windows, index.html is not served at the directory index.
This is a result of the use of `path.join`.

johnwchadwick:
Hello. I would be grateful if you could check this out and hopefully merge it at your earliest convenience, as I make heavy use of webpack-dev-server, and this bug is actually quite difficult to work around if you rely on HTML5-style pushState routing.

Many thanks.

Fixes #54.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -2,7 +2,6 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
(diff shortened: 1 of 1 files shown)


bebraw:
Looks good to me. :+1:

sokra:
Thanks

PR #27, opened 2015-02-11
use res.send instead of res.end to enable etag.

kgdev:
use res.send instead of res.end to enable etag.

PR #59, opened 2016-01-22
use res.end: reverting #27

hzoo:
Revert's https://github.com/webpack/webpack-dev-middleware/pull/27

Assuming https://github.com/webpack/webpack-dev-middleware/issues/58 is correct, I'm just adding a PR to revert.

Otherwise feel free to ignore/close.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -186,7 +186,7 @@ module.exports = function(compiler, options) {
 					res.setHeader(name, options.headers[name]);
 				}
 			}
-			res.send(content);
+			res.end(content);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


middleware.js
```diff
@@ -186,7 +186,7 @@ module.exports = function(compiler, options) {
 					res.setHeader(name, options.headers[name]);
 				}
 			}
-			res.send(content);
+			res.end(content);
```
kaicataldo: What about a guard that defaults to the original implementation when Express isn't used (`karma-webpack` as an example) 
```
if (res.send) res.send(content);
else res.end(content);
```

hzoo: Yeah that could work as well

sokra: :+1:

hzoo: Updated with another commit (let me know if you need me to squash instead)

sokra:
Thanks

hzoo:
Thanks for merging and your work on webpack!

PR #62, opened 2016-02-01
Allow slash in filename regex

koteus:
I started the `webpack-dev-server` in lazy mode with the following arguments `--lazy --iframe --progress --colors`. My webpack.config.js looks like this:

```
'use strict';

var path = require('path');

module.exports = {
    entry: [
        './app/index'
    ],
    output: {
        path: path.join(__dirname, 'build'),
        filename: 'bundle.js',
        publicPath: '/static/'
    }
};
```

I got 404 error for bundle.js. After some debugging it turned out that this expression `(!options.filename || options.filename.test(filename))` on line 150 evaluates to `false` because `filename` variable equals to "/bundle.js" (with a forward slash). So I changed options.filename regex to allow optional forward slash and everything works now correctly.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -24,7 +24,7 @@ module.exports = function(compiler, options) {
 			var str = options.filename
 				.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")
 				.replace(/\\\[[a-z]+\\\]/ig, ".+");
-			options.filename = new RegExp("^" + str + "$");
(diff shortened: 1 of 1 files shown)


koteus:
See this stackoverflow question for more details: http://stackoverflow.com/questions/34960032/what-does-the-lazy-mode-of-webpack-dev-server-do/35132288#35132288

sokra:
Thanks

mathieumg:
I imagine this is only for Webpack 2.x?

SpaceK33z:
Nope, it was released in stable a while ago.

PR #73, opened 2016-03-01
Support Range requests commonly used by media

mzgoddard:
Seeking through media, such as audio and video, often leads a browser
to make range requests, in addition to other scenarios. Responding with
the full content instead of the requested range and can lead to
unexpected behaviour in a web application that in production will
otherwise work since the production server supports range requests.

- Depend on range-parser to parse range request headers
- Use source from the `send` npm package on responding to range
  requests https://github.com/pillarjs/send/blob/31ceb6aaf530152c29a1855bf3f9dbd86533303e/index.js#L551-L583

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -141,6 +141,34 @@ module.exports = function(compiler, options) {
 		return filename ? pathJoin(compiler.outputPath, filename) : compiler.outputPath;
 	}
 
+	function handleRangeHeaders(content, req, res) {
+		if (req.headers['Accept-Ranges']) res.setHeader('Accept-Ranges', 'bytes');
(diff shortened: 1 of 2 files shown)


mzgoddard:
Was writing a little audio test app to test some audio libraries and custom code and found my custom code tripping weirdly. Ended up realizing it was because webpack-dev-middleware was responding with the full content to range requests the browser was making for audio tags.

https://github.com/mzgoddard/html-audio-test

Currently the `user action continuous tag playback` test fails without the range request support if you are interested in looking at my test environment.

sokra:
Thanks

Issue #79, opened 2016-03-24
req.path is sometimes not defined


thatjsguy:
As of https://github.com/webpack/webpack-dev-middleware/pull/78, I am noticing that `req.path` is not always defined. I am using Express 4 here and have found [a little information](http://stackoverflow.com/questions/12525928/how-to-get-request-path-with-express-req-object?answertab=votes#tab-top), but it doesn't explain the *why*. The OP mentions "positioning of middleware", but I am unsure how that might trigger this bug. Overall, I'm wondering what the advantage is of using `req.path` vs `req.url`. Here's the stack trace:

```
TypeError: Cannot read property 'indexOf' of undefined
    at getFilenameFromUrl (/Users/stephenmarshall/code/singular-web-server/node_modules/webpack-dev-middleware/middleware.js:129:9)
    at webpackDevMiddleware (/Users/stephenmarshall/code/singular-web-server/node_modules/webpack-dev-middleware/middleware.js:175:18)
    at call (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:239:7)
    at next (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:183:5)
    at handleIe (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/foxy/lib/utils.js:191:5)
    at call (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:239:7)
    at next (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:183:5)
    at next (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:161:14)
    at Function.handle (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:186:3)
    at Server.app (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/connect/index.js:51:37)
    at Server.<anonymous> (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/socket.io/node_modules/engine.io/lib/server.js:434:22)
    at Server.<anonymous> (/Users/stephenmarshall/code/singular-web-server/node_modules/browser-sync/node_modules/socket.io/lib/index.js:260:16)
[...]

PR #80, opened 2016-03-25
Improve `getFilenameFromUrl` to work with proxy requests

ijse:
This fix issue https://github.com/davezuko/react-redux-starter-kit/issues/666 and as well Express 4 that req.path is undefined in some situation.


jhnns:
Sorry, I've reverted your commit (b05cd045d83123d410c54627dec937506b8013e3) because it broke a lot of code out there (#79 #81 #82).

Could you tell me what the purpose of b05cd045d83123d410c54627dec937506b8013e3 was?

I'm not sure if I'm understanding this PR correctly, but it looks to me that it does not cover the old implementation. I won't merge this PR until its purpose is clear to me.

Sorry again, that we don't have any tests yet. It makes contributing a lot harder. It's certainly not your fault.

ijse:
I use webpack-dev-server and express, and SwitchOmega Plugin to proxy requests to the localhost. so that the request hostname won't match because `options.publicPath='/xxx'`, the `getFilenameFromUrl`will return false.

I wrote this gist to explain the case:

https://gist.github.com/ijse/ffa06369190ebfe4ac11

SpaceK33z:
@ijse, is it not possible to exclude dev-server from the proxy? Also see https://github.com/webpack/webpack-dev-server/issues/600#issuecomment-245868647. Not excluding this seems to result in more issues after this.

SpaceK33z:
I recently added extensive tests for the `getFilenameFromUrl` method, so this will be a lot easier to test.

lib/GetFilenameFromUrl.js
```diff
@@ -1,21 +1,36 @@
...
+	var urlObject = urlParse(url);
+
+	// publicPath has the hostname that is not the same as request url's, should fail
+	if(localPrefix.hostname !== null && urlObject.hostname !== null &&
+		   localPrefix.hostname !== urlObject.hostname) {
+		return false;
+	}
+
+	// publicPath is not in url, so it should fail
+	if(publicPath && localPrefix.hostname === urlObject.hostname && !new RegExp('^' + publicPath).test(url)) {
```
SpaceK33z: Inserting `publicPath` into `RegExp` can lead to unexpected errors, since the string will be [interpreted as regex](https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex).

Maybe use [`lastIndexOf`](https://stackoverflow.com/questions/646628/how-to-check-if-a-string-startswith-another-string#answer-4579228)?

lib/GetFilenameFromUrl.js
```diff
@@ -1,21 +1,36 @@
...
-	if(filename.indexOf("?") >= 0) {
-		filename = filename.substr(0, filename.indexOf("?"));
+
+	// strip localPrefix from the start of url
+	if(urlObject.pathname.indexOf(localPrefix.pathname) === 0) {
+		filename = urlObject.pathname.substr(localPrefix.pathname.length);
+	}
+
+	if(!urlObject.hostname && localPrefix.hostname &&
+	   !new RegExp('^' + localPrefix.path).test(url)) {
```
SpaceK33z: Same with the `RegExp` as above.

lib/GetFilenameFromUrl.js
ijse: `lastIndexOf` or `indexOf` ? 

```
{
    url: "/more/complex/path.js",
    outputPath: "/a",
    publicPath: "/complex"
}
```
Is it expected to be `false`?

SpaceK33z: I think `lastIndexOf`, see [this question](https://stackoverflow.com/questions/646628/how-to-check-if-a-string-startswith-another-string#answer-4579228).

Yes, the example you've given is expected to be `false`, because `publicPath` is not in the url. Would be nice if you add your example to the tests.

SpaceK33z: Ah, just saw your changes with `indexOf`, that's also ok.

SpaceK33z:
One more thing, could you add a testcase that failed before, but is fixed now? Otherwise someone could refactor it again and stop your usecase from working ;).

SpaceK33z:
Thank you!

PR #81, opened 2016-03-25
Fix issue with browsersync

zwhitchcox:
Some middleware and servers only `req.url` and not `req.path`, such as browser-sync and hygienist-middleware. This will serve as a fallback in case req.path is not available.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -171,7 +171,7 @@ module.exports = function(compiler, options) {
 
 	// The middleware function
 	function webpackDevMiddleware(req, res, next) {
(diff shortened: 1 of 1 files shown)


jhnns:
Thank you for your PR.

I've reverted the broken commit. I don't know why @sokra has merged it in the first place. Its purpose is unclear to me.

 I'll try to publish this asap (don't have the rights to publish it on npm :disappointed:).

zwhitchcox:
Ok, thanks Johannes

PR #84, opened 2016-04-11
added readme for native http server, made next optional

nordsimon:
Hi

I'm working on a project now that won't need express. I found that your middleware worked great on native http server and added some readme on how to use without express or any other router. I also made the next callback optional.

Regards
Simon

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,28 @@ app.use(webpackMiddleware(webpack({
 }));
 ```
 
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Sorry, I'm closing this.

At the moment this might happen to work, but it is meant as middleware. If this would be documented, it also means we need to maintain this and make sure it works.

PR #93, opened 2016-05-10
Option to specify output encoding

rdiazv:
I'm working on a legacy project adding some modules built in React. The problem is that the project has charset ISO-8859-1 and webpack-dev-middleware always serve assets with UTF-8 encoding.

Probably there isn't a lot of people with this requirement, but it's nice to have more control over this kind of things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,14 @@ app.use(webpackMiddleware(webpack({
 
 	stats: {
 		colors: true
-	}
+	},
 	// options for formating the statistics
+
+	encodingOptions: {
+		charset: 'UTF-8',
+		include: /.*/
+	}
+	// change encoding if you need something different than UTF-8
 }));
 ```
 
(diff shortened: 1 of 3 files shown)


SpaceK33z:
And is it not an option to convert the project to UTF-8? That wouldn't be very hard, I _think_.

SpaceK33z:
Closed accidentally.

rdiazv:
On this specific case it's not possible. The entire stack it's usign the same encoding and converting all layers isn't that easy. Having said that, I know that this shouldn't be a common problem so it's ok to just close the issue, we are using this branch anyway ;)

SpaceK33z:
Okay, I think it's best to close this PR then, because it's a very specific use-case, and adds an extra dependency. If more people find this an issue, we can always revisit this PR. Thanks anyway for taking the time!

PR #94, opened 2016-05-12
set statusCode to 200 to capable of koa 

MofeLee:
when I use koa to serve webpack-dev-middleware, it will always return 404

because this line
https://github.com/koajs/koa/blob/master/lib/application.js#L133

image
image
image

but body return normally(after rebuild assets, so url is different)
image

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -216,6 +216,7 @@ module.exports = function(compiler, options) {
 					res.setHeader(name, options.headers[name]);
 				}
 			}
+			res.statusCode = 200;
 			if (res.send) res.send(content);
 			else res.end(content);
 		}


MofeLee:
example code:

https://github.com/MofeLee/blog/blob/webpack-hot-middleware-example/webpack/webpack-dev-server.js

`npm run dev`

SpaceK33z:
So in what way is this a problem with this package? Or what do you suggest to do?

SpaceK33z:
Thanks!

Carrotzpc:
@SpaceK33z I encountered the same problem , please update the npm , thanks ~

SpaceK33z:
@Carrotzpc, I want to fix a few other things first, so you'll have to wait for a ~ week.

Carrotzpc:
@SpaceK33z Well , I do a workaroud by [koa-webpack-dev-middleware](https://github.com/yiminghe/koa-webpack-dev-middleware).

PR #91, opened 2016-05-01
Reporter option

Nitive:
Usage:
```js
app.use(webpackDevMiddleware(compiler, {
  reporter: ({ state, stats, options }) => {
    console.log('log what you want')
  }
}))
```
Example reporter:
```js
const oraReporter = () => {
  const spinner = ora()
  spinner.start()
  return ({ state, stats }) => {
    if (state) {
      const durations = stats.endTime - stats.startTime
      const formatedDurations = durations >= 1000 ? `${durations / 1000} s` : `${durations} ms`
      const message = `Completed in ${formatedDurations.magenta}`

      if (stats.hasErrors() || stats.hasWarnings()) {
        spinner.clear()
        const { errors, warnings } = stats.compilation
        errors.concat(warnings).forEach(warning => {
          console.log(warning.message)
        })
      }

      spinner.text = message
    } else {
      spinner.text = 'Building...'
    }
  }
}
```

PR #95, opened 2016-05-13
Support directing log output to custom logger object

maghoff:


SpaceK33z:
Do you still need this now that #91 is merged?

maghoff:
It looks like #91 would solve my problem, yes :)

Do I understand correctly that with #91 merged, webpack would no longer write anything directly to `console.log` but instead go via the reporter? If yes, this definitely solves my problem.

SpaceK33z:
Awesome! Looking at the source code, the only thing that it will still write to console is `webpack: wait until bundle finished:`. This can be disabled using the `noInfo` or `quiet` option though. If you would like this to go through the reporter as well, a PR would be welcome. In that case, let me know, so I can wait with the release.

PR #96, opened 2016-05-16
Only use middleware for 'GET' requests

alexprice91:
Doesn't make sense to for the middleware on all routes.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -172,6 +172,10 @@ module.exports = function(compiler, options) {
 
 	// The middleware function
 	function webpackDevMiddleware(req, res, next) {
+		if(req.method !== 'GET') {
+			return next();
+		}
+		
 		var filename = getFilenameFromUrl(req.url);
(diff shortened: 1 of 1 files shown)


alexprice91:
@sokra looks like updates to `middleware` are somewhat slow. I think this is a pretty usual requirement for this package, even if its just for development. Should I fork this repo and work from there instead?

SpaceK33z:
What is the usecase for this?

alexprice91:
Let's say I have an endpoint GET `/login`.
It takes me to a login form and `webpack-dev-middleware` delivers that.

Then I have another endpoint POST `/login`, this endpoint actually logs a user in.

SpaceK33z:
Thanks!

Small feedback thingy for next time: in the commit title, tell why you've changed something, instead of telling what file; we can already see that in the diff.

SpaceK33z:
Released in `1.7.0`.

PR #98, opened 2016-05-24
Added the ability to use historyApiFallback

Kikketer:
Give users the ability to use the historyApiFallback similar to what's in the webpack.config file.  This simply uses the `stat` exception that's thrown and then uses the sent in `index` to render the index page.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# @AngularClass
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+insert_final_newline = false
+trim_trailing_whitespace = false
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.idea
--- a/README.md
+++ b/README.md
@@ -64,6 +64,11 @@ app.use(webpackMiddleware(webpack({
 		colors: true
 	}
 	// options for formating the statistics
+	
+	historyApiFallback: {
+		index: "/"
+	}
+	// similar to webpack.config for historyApiFallback
 }));
 ```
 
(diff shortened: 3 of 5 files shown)


kripod:
@sokra Please merge this or #101 as soon as possible, as this would greatly improve our koa development workflow.

Kikketer:
Any updates on pull requests?  Is there something more you need before we can get this in place?

SpaceK33z:
Is it not possible to do something [like this](https://github.com/webpack/webpack-dev-middleware/pull/44#issuecomment-170462282)?

SpaceK33z:
I'm closing this issue because of inactivity. Feel free to comment, and we can always re-open it again.

rgigger:
Is there an offical correct way to do this? Everything I've tried so var has failed.

shellscape:
@rgigger it depends on your server setup and middleware of choice, but you add/use the midleware in this order:

1. webpack-dev-middleware
2. static file middleware
3. historyApiFallback

You can reference webpack-dev-server as well.

donaldpipowitch:
@shellscape Thanks. Do you know a public example of this? I basically use 

1. webpack-dev-middleware
2. webpack-hot-middleware
3. static file middleware
4. historyApiFallback

But when I refresh the browser with a non-root I get 404s 🤔

shellscape:
@donaldpipowitch if you can get away with only supporting browsers which support websockets, I'd give [webpack-hot-client](webpack-contrib/webpack-hot-client) a look. 

Which static file middleware are you leveraging? And express, koa, raw http, etc?

donaldpipowitch:
Thank you, I'll give it a try. I use express. (See [here](https://github.com/Mercateo/ws/blob/master/src/lib/express.ts#L16).)

shellscape:
Ah, so I just checked my test app harness for `hot-client` and I was mistaken, at least in the context of koa's `koa-static` module... I `use` the static middleware _before_ the dev-middleware. See: https://github.com/webpack-contrib/webpack-hot-client/blob/master/test/setup.js#L48. I added a subdirectory resource to my manual test app in that repo to confirm that it allows non-root static files and it's all good. In theory, `express.static` should work the same.

PR #100, opened 2016-05-30
Removing watching/watching.running check

### LINKED ISSUE #107: waitUntilValid callback always called immediately when using MultiCompiler.
It seems that this snippet always hits the first code path when using a compiler built with an array of configurations, because watching is an instance of [MultiCompiler](https://github.com/webpack/webpack/blob/master/lib/MultiCompiler.js#L9) which doesn't have a running property.

```javascript
webpackDevMiddleware.waitUntilValid = function(callback) {
  callback = callback || function(){};
  if (!watching || !watching.running) callback();
  else ready(callback, {});
};
```

ibash:
Given that `state` should always have the correct value (if the bundle is valid
or not) then we should be able to use ready directly without checking the
watcher.

I also hit a bug where the watcher had running set to false while the bundle was
invalid, leading to a callback being executed immediately instead of waiting.

Fixes #107

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -265,8 +265,7 @@ module.exports = function(compiler, options) {
 
 	webpackDevMiddleware.waitUntilValid = function(callback) {
 		callback = callback || function() {};
-		if(!watching || !watching.running) callback();
-		else ready(callback, {});
+		ready(callback, {});
 	};
 
 	webpackDevMiddleware.invalidate = function(callback) {


SpaceK33z:
@ibash, PR seems good. Could you verify that it still is necessary (it's been a while), and if so, fix the merge conflict? Will merge after that.

ibash:
Fixed the merge conflict. I'm not using this change myself anymore, but looking over it again I think it is still valid.

SpaceK33z:
Thank you!

PR #101, opened 2016-06-09
Added support for the `historyApiFallback` option

kripod:


kripod:
It turns out that this solution breaks hot module replacement, because it even ovverrides the HMR path...

PR #110, opened 2016-07-30
readme.md change: publicPath should be stated as required

torioLuz:
publicPath should be stated as required, as also described on the webpack docs: [https://webpack.github.io/docs/webpack-dev-middleware.html](url)

I think this can avoid confusion for beginners.


SpaceK33z:
Thanks!

PR #111, opened 2016-08-03
added support for koa middleware + some refactoring for the code

NXTaar:
Added support for koa applications (generator based `koa v1.2~`). By default, it uses the middleware for Express. If there is the `koa` flag in the options set to `true` it will use generator based middleware for `koa`.
Also did some refactoring for better code readability.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -2,6 +2,7 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
+'use strict';
 var MemoryFileSystem = require("memory-fs");
 var mime = require("mime");
 var parseRange = require("range-parser");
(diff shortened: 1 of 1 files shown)


SpaceK33z:
This would require Node v4+. I'm also wondering that if we want to support Koa, it's maybe better to use a separated file for this. This looks very hacky.

SpaceK33z:
@NXTaar, you still interested in fixing this?

grawk:
This PR is interesting to me, because it does attempt to refactor the module to be useful in multiple contexts. I'm currently working on a "construx" module for webpack and finding this module as a great starting point (for reference, [construx](https://github.com/krakenjs/construx)).

Because the main "construx" module furnishes the middleware, it would call out to my `construx-webpack` module only to access the memory-fs. Therefore, I need to remove the references to req/res/next from this module's code in order to do that.

I haven't thought through it completely yet, but it seems it should be possible to refactor this module such that the essential part of it (pass a file name, get the bundled file content in the callback) can be pulled up into an exportable function or separate module which this module can depend on.

Anyway, just throwing that out there as food for thought.

SpaceK33z:
@grawk, I'm open to refactoring this package so that it is easier to add support for other servers (e.g. Koa) or plugins.

I don't have time for this myself currently, so if anyone is willing to take a stab at it: I'll happily review a PR. You could even do this without adding Koa support; that could be done in a second PR.

Note though that there are integration tests now, so it's important to update these, and preferably add more unit tests.

CrocoDillon:
Hey! I'm interested to see how this PR progresses and am happy to help if needed.

Perhaps a separate file is more interesting if requiring a higher version of Node than previously for connect users is an issue. Something like `import webpackDevMiddleware from 'webpack-dev-middleware/koa'` could work.

Here's the thing, Koa 2 will be different. So you'd also need `import webpackDevMiddleware from 'webpack-dev-middleware/koa2'` or something.

NXTaar:
@SpaceK33z The #140 is good, i suggest to resolve and merge it first, and then i will make a new PR to merge my koa-supporting changes. I want to avoid to doPR's to @grawk's fork. Better we all improve the original :-)

PR #116, opened 2016-08-31
Fix #113, drop extraneous condition

### LINKED ISSUE #113: Range support looks like it has issues
The range support doesn't look right:

https://github.com/webpack/webpack-dev-middleware/blob/264dd2bc1810b4ce8a5647c90076433e0da7e8bc/middleware.js#L169
- don't think there is such a **request** header "Accept-Ranges"
- either way, it won't ever match, node.js req headers are always lower cased

[update] updated link to point to correct line to latest commit instead of master

hnry:
The if condition will never be met so drop it.
Calls to handleRangeHeaders() means the resource can be byte served, so
always send back response header indicating so.


SpaceK33z:
Thanks!

hnry:
ping @SpaceK33z :( commit also accidentally included my ".DS_Store" file, sorry I should've double checked.

Would you like to correct it yourself? Or is it better for you to revert, and I re-pull?

SpaceK33z:
Ahh completely missed that too. Will fix it myself soon.
On Sep 1, 2016 18:51, hnry <notifications@github.com> wrote:ping @SpaceK33z :( commit also accidentally included my ".DS_Store" file, sorry I should've double checked.

Would you like to correct it yourself? Or is it better for you to revert, and I re-pull?

SpaceK33z:
Released in `1.7.0`.

PR #124, opened 2016-09-16
Make log & warn functions replaceable

alanpearce:
By allowing overrides of the functions used for logging, we can allow the output from this middleware to be processed according to the logging library used in the application.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -54,7 +54,7 @@ module.exports = function(compiler, options) {
 	}
 	if(typeof options.reporter !== "function") options.reporter = defaultReporter;
 	if(typeof options.log !== "function") options.log = console.log.bind(console);
-	if(typeof options.warn !== "function") option.warn = console.warn.bind(console);
+	if(typeof options.warn !== "function") options.warn = console.warn.bind(console);
 
 	// store our files in memory
 	var fs;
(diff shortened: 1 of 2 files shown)


SpaceK33z:
There is also a `reporter` option, that does about the same. There is only one `console.log` that is not inside this reporter. Could you maybe move that `console.log` over there instead?

alanpearce:
Well, the reason I didn't touch `reporter` is because it seems non-trivial to replace as a library user, and then relies on the API not breaking over time. Also, from the name, I assumed it would be a function run at the end of a compilation to format the output. IMO it should not even be used to log that the bundle is invalid, just to report the stats (or not) from a compilation.

I could do it, but it seems incorrect to me. I'd have to add some other `state` value or some other decision flag for the other `console.log`, and then what happens if maybe more logging is added later? More `state` values? Would it really be desirable to maintain a most-likely increasing number of switch cases rather than adding a log line, which could then be logged in a way that fits into whatever framework or log library the user may be using?

If you really think it should be done like this, please let me know what you'd like to have as the `state` value for the last log call outside of the reporter, or how to otherwise decide if this is a "waiting" event.

SpaceK33z:
Okay, fair enough. We can let the `reporter` method only be for processing the output.

Could you add a test for this and fix the broken tests? The relevant tests are in `test/Reporter.test.js`.

SpaceK33z:
Thanks!

PR #129, opened 2016-09-20
Fix protocol-relative urls with `publicPath`

### LINKED ISSUE #128: Webpack dev server does not serve any assets
## Environment details:

Most of our workstations are on node `6.3.0` and `6.5.0`. And npm `3.10.3`.

# Issue

Seems like a recent change in our `npm-shrinkwrap.json` broke the webpack-dev-server (seems like the server cannot find / server any bundle).

We were able to figure out that bumping `webpack-dev-middleware` from `1.6.1` to `1.8.1` bricks the dev server. While pinning down the middle to `1.6.1` unbricks the server.

Is it normal to specify the `webpack-dev-middleware` dev dependency in our `package.json`? Should we try to avoid it to avoid any future headaches?
Is this a known issue and is there an official fix?
Any help would be much appreciated.

-
Apologies if this is not the right channel to report this. I would appreciate a re-route / suggestions.

razh:
Pass `true` for `slashesDenoteHost` in `url.parse`.

https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost

Fixes #128

### DIFF
--- a/lib/GetFilenameFromUrl.js
+++ b/lib/GetFilenameFromUrl.js
@@ -5,7 +5,7 @@ function getFilenameFromUrl(publicPath, outputPath, url) {
 	var filename;
 
 	// localPrefix is the folder our bundle should be in
-	var localPrefix = urlParse(publicPath || "/");
+	var localPrefix = urlParse(publicPath || "/", false, true);
 	var urlObject = urlParse(url);
 
 	// publicPath has the hostname that is not the same as request url's, should fail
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Clean fix, thank you!

parshap:
Can this change get published to npm please?

rmurphey:
What's the prognosis for getting this released? I was just in the process of opening an issue about this when I discovered it was already fixed on master.

PR #131, opened 2016-09-23
Set default reporter, log, and warn

chrisj-porch:
options.warn is called before a default is set if options.watchDelay is a thing.


SpaceK33z:
Thanks!

PR #135, opened 2016-09-28
Support HTML5 pushState

tsgautier:


SpaceK33z:
Thanks for your PR. This is out of the scope of webpack-dev-middleware however. webpack-dev-server also has support for pushState (also called "historyApiFallback"), which uses the powerful [connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback) package for this.

This seems to be a simple addition, but if we would implement this, later people also would want to customize the `/index.html`. Then rewrite specific urls, etc.

See also [lib/Server.js](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L223-L226) in webpack-dev-server.

tsgautier:
I understand your position.  

However I don't respectfully disagree with where you've decided to draw the line - if I am using webpack-dev-middleware to develop in a development mode, I can do so for apps that don't require pushState.  However, if I am developing an app that requires pushState, I cannot use it - hence my PR.

I totally get that webpack-dev-middleware should not try to be webpack-dev-server, so I respect your reluctance to add features to it that do so, but it seems that without pushState support, it's only 1/2 able to serve it's purpose (does support apps that don't require pushState, but does not support apps that do).

SpaceK33z:
> However, if I am developing an app that requires pushState, I cannot use it - hence my PR.

But what is letting you from adding the `connect-history-api-fallback` package as middleware? It adds two lines of code (the `require` statement and `app.use()`).

SpaceK33z:
Sorry, but I'm closing this. Next time it would be better to first open an issue for something like this, to prevent wasting time. I don't want to demotivate you from contributing, the code itself is good.

PR #140, opened 2016-10-13
factor out shareable methods to a common class

grawk:
Leaving this here to get your impression of this approach. Using it, it would be fairly simple to add an additional interface to the module. One which simply takes a file name and returns the bundle content.

### DIFF
--- a/lib/Shared.js
+++ b/lib/Shared.js
@@ -1,0 +1,214 @@
+var parseRange = require("range-parser");
+var pathIsAbsolute = require("path-is-absolute");
+var MemoryFileSystem = require("memory-fs");
+var HASH_REGEXP = /[0-9a-f]{10,}/;
+
+module.exports = function Shared(context) {
+	var share = {
+		setOptions: function(options) {
+			if(!options) options = {};
+			if(typeof options.watchOptions === "undefined") options.watchOptions = {};
+			if(typeof options.reporter !== "function") options.reporter = share.defaultReporter;
+			if(typeof options.log !== "function") options.log = console.log.bind(console);
+			if(typeof options.warn !== "function") options.warn = console.warn.bind(console);
+			if(typeof options.watchDelay !== "undefined") {
+				// TODO remove this in next major version
+				options.warn("options.watchDelay is deprecated: Use 'options.watchOptions.aggregateTimeout' instead");
+				options.watchOptions.aggregateTimeout = options.watchDelay;
+			}
+			if(typeof options.watchOptions.aggregateTimeout === "undefined") options.watchOptions.aggregateTimeout = 200;
+			if(typeof options.stats === "undefined") options.stats = {};
+			if(!options.stats.context) options.stats.context = process.cwd();
+			if(options.lazy) {
+				if(typeof options.filename === "string") {
+					var str = options.filename
+						.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&")
+						.replace(/\\\[[a-z]+\\\]/ig, ".+");
+					options.filename = new RegExp("^[\/]{0,1}" + str + "$");
+				}
+			}
+			context.options = options;
+		},
+		defaultReporter: function(reporterOptions) {
+			var state = reporterOptions.state;
+			var stats = reporterOptions.stats;
+			var options = reporterOptions.options;
+
+			if(state) {
+				var displayStats = (!options.quiet && options.stats !== false);
+				if(displayStats && !(stats.hasErrors() || stats.hasWarnings()) &&
+					options.noInfo)
+					displayStats = false;
+				if(displayStats) {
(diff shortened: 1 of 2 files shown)


grawk:
FYI, here is what adding the "direct" interface looks like:
https://github.com/grawk/webpack-dev-middleware/blob/feature.new.api/middleware.js#L104-L156

SpaceK33z:
I'll try it out this weekend. Thanks for your effort!

grawk:
Hi @SpaceK33z. Please let me know if you have any concerns with this PR which I can address. Thanks.

SpaceK33z:
@grawk, sorry, I'm a bit busier than expected. I still have it in my list to review.

lib/Shared.js
```diff
@@ -0,0 +1,197 @@
+var parseRange = require("range-parser");
+var pathIsAbsolute = require("path-is-absolute");
+var MemoryFileSystem = require("memory-fs");
+
+module.exports = function shared(context) {
+	var shared = {
```
shellscape: I've never been a fan of vars sharing the same name as the fn that contains them. Makes for some funky reading. :)

SpaceK33z: Agreed.

grawk: Would you like me to underbar the variable as `_shared` or un-name the function? Or some other resolution?

middleware.js
```diff
@@ -202,31 +39,32 @@ module.exports = function(compiler, options) {
...
+		var filename = getFilenameFromUrl(context.options.publicPath, context.compiler.outputPath, req.url);
 		if(filename === false) return goNext();
 
 		// in lazy mode, rebuild on bundle request
-		if(options.lazy && (!options.filename || options.filename.test(filename)))
-			rebuild();
+		if(context.options.lazy && (!context.options.filename || context.options.filename.test(filename)))
+			shared.rebuild();
 
 		if(HASH_REGEXP.test(filename)) {
```
SpaceK33z: Could this be moved to the shared part completely? I feel like this should not be exposed.

middleware.js
```diff
@@ -202,31 +39,32 @@ module.exports = function(compiler, options) {
...
 		}
 
-		var filename = getFilenameFromUrl(options.publicPath, compiler.outputPath, req.url);
+		var filename = getFilenameFromUrl(context.options.publicPath, context.compiler.outputPath, req.url);
 		if(filename === false) return goNext();
 
 		// in lazy mode, rebuild on bundle request
-		if(options.lazy && (!options.filename || options.filename.test(filename)))
-			rebuild();
+		if(context.options.lazy && (!context.options.filename || context.options.filename.test(filename)))
```
SpaceK33z: Maybe also make a method for this in the shared file?

lib/Shared.js
```diff
@@ -0,0 +1,197 @@
...
+					options.log(stats.toString(options.stats));
+				}
+				if(!options.noInfo && !options.quiet) {
+					options.log("webpack: bundle is now VALID.");
+				}
+			} else {
+				options.log("webpack: bundle is now INVALID.");
+			}
+		},
+		handleRangeHeaders: function handleRangeHeaders(content, req, res) {
```
SpaceK33z: This method assumes that `res.setHeader`, `req.headers.range` and `res.statusCode` work, but we don't know what server will be used, so this is a bit of a leaky abstraction. Don't know the correct way to do this though.

grawk: I added a comment in the method, suggesting a way forward. E.g. conditional(s) based on an object check for various APIs. A new context flag could also be added to indicate what API is being used. But I believe that can be addressed in a future PR for support of additional server(s).

SpaceK33z: All right, fair enough.

SpaceK33z:
@grawk, could you sign the CLA?

grawk:
@SpaceK33z Hm. Somehow it seems these commits got associated with my corporate github ID. Therefore when I authorized the CLA app with my public github account, it's not able to link that with this changeset. Not sure how that happened :( Will look into it.

grawk:
I've associated my other email address (found in .gitconfig), medelman@paypal.com, with my "grawk" account. It's not seemingly resolving the issue as far as the CLA app is concerned. Is there some other way I can verify to you I've signed the CLA or what resolution would you recommend?

SpaceK33z:
I think squashing or making a new PR with the correct account is best. Sorry for the trouble.

grawk:
OK. I was afraid you'd say that 😆 

I'll probably have to table this today and complete it tomorrow.

grawk:
squashed into a single commit, although the CLA badge is not reflecting a different status

SpaceK33z:
Thanks, awesome work. I'll probably tweak some things here and there, but it looks good :).

grawk:
Thank you :) Happy to help.

SpaceK33z:
FYI, Published this in `1.9.0`.

PR #150, opened 2016-11-22
Add a option, mimeTypes, to support defining custom mime types

xiaoyann:
Please look here https://github.com/webpack/webpack-dev-middleware/blob/master/middleware.js#L65:

```js
res.setHeader("Content-Type", mime.lookup(filename) + "; charset=UTF-8");
```

The line above detemine the mimeType of file, but I have a special file, Index.phtml, so I have to defining custom mimeType

Example:
```js
app.use(webpackMiddleware(webpack({
    // webpack options
    // webpackMiddleware takes a Compiler object as first parameter
    // which is returned by webpack(...) without callback.
    entry: "...",
    output: {
        path: "/"
        // no real path is required, just pass "/"
        // but it will work with other paths too.
    }
}), {
    // ......

    mimeTypes: {
        'text/html': ['phtml']
    }
}));
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,11 @@ app.use(webpackMiddleware(webpack({
 	headers: { "X-Custom-Header": "yes" },
 	// custom headers
 
+	mimeTypes: { "text/html": [ "phtml" ] },
+	// Add custom mime/extension mappings
+	// https://github.com/broofa/node-mime#mimedefine
+	// https://github.com/webpack/webpack-dev-middleware/pull/150
+
 	stats: {
 		colors: true
 	}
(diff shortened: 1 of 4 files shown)


SpaceK33z:
Please explain your use case.

SpaceK33z:
I seem to have looked over your use case. Could you document this in the README?

xiaoyann:
👌

xiaoyann:
I get it. Thank you.

SpaceK33z:
Hey, I'm planning to release a new version somewhere next week. Do you think you have time to fix this next week? It's no problem if you don't have the time, but just want to know this so I can wait for it ;).

shellscape:
@xiaoyann ping! please see https://github.com/webpack/webpack-dev-middleware/pull/150#pullrequestreview-11779034. would you be able to add a test for this change and move the change to the shared file as requested?

xiaoyann:
@SpaceK33z I have moved the change to the Shared.js and added test.

PR #157, opened 2016-12-10
Fix error handling in compiler `watch()` and `run()`

### LINKED ISSUE #156: Errors in async tasks cause webpack to hang
I want to report a **bug**.

## What is the current behavior?
Webpack hangs indefinitely when using `webpack-dev-middleware` (directly or indirectly through `webpack-dev-server`) and there's an error in an async step.

## Steps to reproduce.
Make a plugin like this:

```typescript
class MyErrorPlugin {
  apply(compiler) {
    compiler.plugin('make', (compilation, cb) => {
      setTimeout(() => cb(new Error('Heho!')), 1);
    });
  }
}
```

Run a webpack configuration with `webpack-dev-middleware` and this plugin, and the generation will hang forever.

Note that this occurs in the Webpack plugin for Angular 2, when an application cannot be statically analyzable, since we perform that analysis . Also, if this error is not reported in the async task everything works as the expected behaviour.

The reason is the callback used when running webpack (see [here](https://github.com/webpack/webpack-dev-middleware/blob/6b49341db9c1323a1304ff99092036a72d2449a8/lib/Shared.js#L164) and [here (watch)](https://github.com/webpack/webpack-dev-middleware/blob/6b49341db9c1323a1304ff99092036a72d2449a8/lib/Shared.js#L153)) which simply `throw` instead of reporting the error and stopping the build. Because the build is not stopped by Webpack itself, webpack is waiting for the `make` step of the compiler to finish, which won't happen because the `throw` killed the task.

Not sure what the proper solution is.

# What is the expected behavior?
Show error on console, say the bundle is INVALID or something, watch for file changes, rebuild when file changes.

SpaceK33z:

Bugfix

Yes.

**Summary**

See #156. When the compiler wanted to report an error in the bundle (e.g. because of a plugin), it would throw the error and quit the process.

I've changed this to work more like the [webpack watch mode](https://github.com/webpack/webpack/blob/287eb18076d80688be03a55431918b628b25f436/bin/webpack.js#L302-L303); it shows the error with `console.error` but it keeps running (because the error can be fixed with a new compilation).

Optionally, it's possible to hide / customize this error by using `options.error`.

Not really; previously the process would quit and now it doesn't.

**Other information**

Fixes #156

### DIFF
--- a/lib/Shared.js
+++ b/lib/Shared.js
@@ -11,6 +11,7 @@ module.exports = function Shared(context) {
 			if(typeof options.reporter !== "function") options.reporter = share.defaultReporter;
 			if(typeof options.log !== "function") options.log = console.log.bind(console);
 			if(typeof options.warn !== "function") options.warn = console.warn.bind(console);
+			if(typeof options.error !== "function") options.error = console.error.bind(console);
 			if(typeof options.watchDelay !== "undefined") {
 				// TODO remove this in next major version
 				options.warn("options.watchDelay is deprecated: Use 'options.watchOptions.aggregateTimeout' instead");
@@ -149,9 +150,7 @@ module.exports = function Shared(context) {
 			var compiler = context.compiler;
 			// start watching
 			if(!options.lazy) {
-				var watching = compiler.watch(options.watchOptions, function(err) {
-					if(err) throw err;
-				});
+				var watching = compiler.watch(options.watchOptions, share.handleCompilerCallback);
 				context.watching = watching;
 			} else {
 				context.state = true;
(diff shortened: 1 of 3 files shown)


hansl:
This PR doesn't work. I'm still looking at why, but there's still nothing written on the console.

hansl:
Never mind my last message.

This PR works, partially. The error message shows up in the console properly, but the compilation still hangs. In this example I `throw new Error('hello world')`:

```
🏠  ng serve
** NG Live Development Server is running on http://localhost:4200. **
 10% building modules 2/2 modules 0 activeError: hello world
    at /Users/hans/Sources/angular-cli/packages/@ngtools/webpack/src/plugin.ts:235:19
    at process._tickCallback (internal/process/next_tick.js:103:7)
 69% building modules 678/679 modules 1 active ...ns/Sources/_Temp/my-project/src asyncwebpack: wait until bundle finished: /worker.js
```

hansl:
LGTM.

SpaceK33z:
Thanks for verifying @hansl.

PR #164, opened 2016-12-23
Improve messages when bundle is compiling and done with compiling

SpaceK33z:

Change, a small improvement in DX.

Yes

**Summary**

When the bundle is compiling, the console showed this:

```
webpack: bundle is now INVALID.
```

When it is done compiling, regardless of errors, it showed this:

```
webpack: bundle is now VALID.
```

This is very confusing, because most people think when it is "valid", that the bundle has successfully compiled. Also, when it says "bundle is now INVALID", people can think they have done something wrong.

<hr>

This PR indents to clarify these messages;

When the bundle is compiling:

```
webpack: Compiling...
```

When the bundle is done compiling, but has errors:

```
webpack: Failed to compile.
```

When the bundle is done compiling, no errors:

```
webpack: Compiled successfully.
```

Note that if there are warnings, it will say "Compiled successfully." This is because a bundle with warnings will still work. I could add a special case for this, maybe "Compiled successfully, but with warnings." ?

Not really, but I _think_ there might be people who have hacked something based on these messages.

**Other information**

Inspired by [this discussion on Twitter](https://twitter.com/dan_abramov/status/812058255753183232). The messages are also stolen from create-react-app.

cc @gaearon

### DIFF
--- a/lib/Shared.js
+++ b/lib/Shared.js
@@ -44,7 +44,13 @@ module.exports = function Shared(context) {
 					options.log(stats.toString(options.stats));
 				}
 				if(!options.noInfo && !options.quiet) {
-					options.log("webpack: " + (stats.hasErrors() ? "Failed to compile." : "Compiled successfully."));
+					var msg = "Compiled successfully.";
+					if(stats.hasErrors()) {
+						msg = "Failed to compile.";
+					} else if(stats.hasWarnings()) {
+						msg = "Compiled with warnings.";
+					}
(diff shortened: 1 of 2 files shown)


gaearon:
"Compiled with warnings" is a good wording IMO.

SpaceK33z:
Thanks for the help here and on Twitter @gaearon!

gaearon:
Thanks for caring.

PR #173, opened 2017-02-27
Fix parsing urls with spaces. Use decodeURIComponent for getFilenameFromUrl final output

dimapaloskin:

Bugfix.

Yes.

**Summary**

This PR fix getFilenameFromUrl output result. 
getFilenameFromUrl function does not properly parse url with spaces.
It is happens because url.parse return encoded pathname.
Look screenshot please:

image

So getFilenameFromUrl can't return right filename that contains spaces.
My approach is decode final output again.

It's shouldn't.

### DIFF
--- a/lib/GetFilenameFromUrl.js
+++ b/lib/GetFilenameFromUrl.js
@@ -3,7 +3,6 @@ var urlParse = require("url").parse;
 
 function getFilenameFromUrl(publicPath, outputPath, url) {
 	var filename;
-	url = decodeURIComponent(url);
 
 	// localPrefix is the folder our bundle should be in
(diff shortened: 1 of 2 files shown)


SpaceK33z:
On this line, https://github.com/dimapaloskin/webpack-dev-middleware/blob/33c244851f70ffc308658b6c530beb58f811835a/lib/GetFilenameFromUrl.js#L6, we already decode the URI. Perhaps that decode function should be removed now?

dimapaloskin:
@SpaceK33z Done.

SpaceK33z:
Thanks!

Issue #180, opened 2017-04-12
Feature: Fallback for unreachable file


lo-tp:
Imagine, we are debugging a react app with react-router. We click
a link sending us to another url. That's ok since all the routing
stuff are handled in the front end. Then we make some modification
to the html and want to refresh the whole page. Now problem arises,
the url is now unreachable in the back end because we have changed
it so that the server has no idea what file to serve.

In this case, we would like the dev server to serve one similiar
html no matter what the url is( the root html).

shellscape:
Closed considering resolution of #181

PR #181, opened 2017-04-12
Add a fallback argument to handle unreachable file

lo-tp:
#180
Imagine, we are debugging a react app with react-router. We click
a link sending us to another url. That's ok since all the routing
stuff are handled in the front end. Then we make some modification
to the html and want to refresh the whole page. Now problem arises,
the url is now unreachable in the back end because we have changed
it so that the server has no idea what file to serve.

In this case, we would like the dev server to serve one similar
html no matter what the url is( the root html). That would be the behavior
when fallBack argument is set to true after this commit is merged.

I added a new feature.

Yes

No. The default behavior is untouched if the fallBack argument is
not set to true.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -22,7 +22,7 @@ module.exports = function(compiler, options) {
 	var shared = Shared(context);
 
 
-  // The middleware function
+        // The middleware function
 	function webpackDevMiddleware(req, res, next) {
 		function goNext() {
 			if(!context.options.serverSideRender) return next();
@@ -61,7 +61,7 @@ module.exports = function(compiler, options) {
 				return goNext();
 			}
 
-      // server content
+                        // server content
 			var content = context.fs.readFileSync(filename);
 			content = shared.handleRangeHeaders(content, req, res);
 			res.setHeader("Access-Control-Allow-Origin", "*"); // To support XHR, etc.
@@ -72,7 +72,7 @@ module.exports = function(compiler, options) {
 					res.setHeader(name, context.options.headers[name]);
 				}
 			}
-      // Express automatically sets the statusCode to 200, but not all servers do (Koa).
+                        // Express automatically sets the statusCode to 200, but not all servers do (Koa).
 			res.statusCode = res.statusCode || 200;
 			if(res.send) res.send(content);
 			else res.end(content);
--- a/middleware.js
+++ b/middleware.js
@@ -1,7 +1,7 @@
 /*
-   MIT License http://www.opensource.org/licenses/mit-license.php
-   Author Tobias Koppers @sokra
-   */
+ MIT License http://www.opensource.org/licenses/mit-license.php
+ Author Tobias Koppers @sokra
+ */
 var mime = require("mime");
 var getFilenameFromUrl = require("./lib/GetFilenameFromUrl");
 var Shared = require("./lib/Shared");
@@ -22,7 +22,7 @@ module.exports = function(compiler, options) {
 	var shared = Shared(context);
 
 
-        // The middleware function
+	// The middleware function
 	function webpackDevMiddleware(req, res, next) {
 		function goNext() {
 			if(!context.options.serverSideRender) return next();
@@ -61,7 +61,7 @@ module.exports = function(compiler, options) {
 				return goNext();
 			}
 
-                        // server content
+			// server content
 			var content = context.fs.readFileSync(filename);
 			content = shared.handleRangeHeaders(content, req, res);
 			res.setHeader("Access-Control-Allow-Origin", "*"); // To support XHR, etc.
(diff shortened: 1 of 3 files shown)


middleware.js
```diff
@@ -22,7 +22,7 @@ module.exports = function(compiler, options) {
 	var shared = Shared(context);
 
 
-	// The middleware function
```
shellscape: please don't change whitespace formatting :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webpack-dev-middleware",
-  "version": "1.10.1",
+  "version": "1.10.2",
```
shellscape: please don't update package versions, we'll do that on our end for publish

test/API.test.js
```diff
@@ -1,5 +1,6 @@
 var middleware = require("../middleware");
 var should = require("should");
+require("mocha-sinon");
```
shellscape: please move this to line 1 and add a line break between it and `var middleware...`

shellscape:
@lo-tp thanks for cleaning up those spots. looks like there's a few conflicts to resolve. I'd recommend rebasing

lo-tp:
@shellscape Done rebasing.

lo-tp:
Hi, @shellscape , I have done rebasing, is there any other thing I can do to speed up the merging process?

middleware.js
shellscape: @lo-tp please revert the whitespace changes

middleware.js
```diff
@@ -40,6 +40,9 @@ module.exports = function(compiler, options) {
 		if(filename === false) return goNext();
 
 
+		if(!context.fs.existsSync(filename) && context.options.fallBack) {
+			filename = getFilenameFromUrl(context.options.publicPath, context.compiler.outputPath,context.options.publicPath);
```
shellscape: inconsistent spaces after commas. it's a nitpick but it's good to be consistent

shellscape:
@lo-tp looks like your editor is setup for some specific whitespace settings and it may be automatically changing things. please revert the whitespace changes and clean up spacing between commas (i'll look at improving the linting in the future). 

other than that the feature looks good. comparing for a truthy value of `options.fallback` will allow for furthering this feature in the future if needed.

lo-tp:
@shellscape . My fault, I'll make sure that all the modifications I made are consistent with the eslint configuration in the code base.

shellscape:
@lo-tp thanks for making those changes. I was discussing the PR offline and was asked about the  option in webpack-dev-server. https://webpack.github.io/docs/webpack-dev-server.html#the-historyapifallback-option. So we have to ask - can you use this option instead of the feature you've implemented in this PR? We want to make sure we're not duplicating functionality.

(Folks that don't use webpack-dev-server (who are using express, koa, etc) Can use https://github.com/bripkens/connect-history-api-fallback, which is the same middleware that webpack-dev-server uses for the option.)

lo-tp:
@shellscape , I am so happy to see that there is an existing solution to my problem.
With https://github.com/bripkens/connect-history-api-fallback, every thing works so smoothly.
Think you for telling me this package.

shellscape:
Wonderful! Thanks for all of your work on this PR, and I'm sorry I wasn't able to point you at that module sooner.

kenjura:
I don't understand how connect-history-api-fallback is a solution to this problem. Perhaps I'm not implementing it correctly. Here is the scenario:

In local dev, I start my project by running index.js. This activates an Express backend, then kicks in webpack-dev-middleware to compile the frontend and serve it.

Once ready, requests may go to myapp.com/, but also myapp.com/foo/bar or myapp.com/foo/123

In the former case, webpack-dev-middleware serves the frontend bundle as expected, and history API changes work as expected. However, the local dev server should also respond to the latter two URLs, serving the exact same frontend bundle, as would happen with the fallBack feature in this PR.

The problem with connect-history-api-fallback seems to be that it is assuming the file exists on the local file system. By design, webpack-dev-middleware creates an in-memory file system. The index.html and all bundled assets reside there--not on the file system--and therecore connect-history-api-fallback cannot serve them.

In my experience, SPAs typically operate from a single root HTML file with all necessary JS to render all possible routes, and although they use a frontend router and the History API to navigate within a session, their web servers should also respond to deep links exactly the same as if the user started from the index page and navigated using History.

In order for that to work with webpack-dev-middleware, it would seem the feature in this PR is required.

Am I not understanding something? Is there some other solution that involves webpack-dev-middleware? I was previously using webpack-dev-server plus a separate process for the backend, but I thought the point of this middleware was that I could unify the two.

AlexKvazos:
@kenjura Mentioned something very important. Currently I see no way of using webpack-dev-middleware to serve SPA without having to request the base url every refresh.

PR #186, opened 2017-04-18
Use options.log, warn and error for compilation messages

Tarrask:
When the compilation fails or has warnings, the compilation
messages are outputted in console.log not using the options.error
and options.warn functions.

This commit change the defaultReporter to output the stats
and final message to the options.error when the compilation
has errors or the options.warn when it has warnings.

The Reporter.test.js test cases are updated accordingly.

### DIFF
--- a/lib/Shared.js
+++ b/lib/Shared.js
@@ -41,16 +41,22 @@ module.exports = function Shared(context) {
 					options.noInfo)
 					displayStats = false;
 				if(displayStats) {
-					options.log(stats.toString(options.stats));
+					if(stats.hasErrors()) {
+						options.error(stats.toString(options.stats));
+					} else if(stats.hasWarnings()) {
+						options.warn(stats.toString(options.stats));
+					} else {
+						options.log(stats.toString(options.stats));
+					}
 				}
 				if(!options.noInfo && !options.quiet) {
-					var msg = "Compiled successfully.";
 					if(stats.hasErrors()) {
-						msg = "Failed to compile.";
+						options.error("webpack: Failed to compile.");
 					} else if(stats.hasWarnings()) {
-						msg = "Compiled with warnings.";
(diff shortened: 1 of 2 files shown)


Tarrask:
Hi jsf-clabot, can you fix the page you ask me to sign. The link to the licence file is not working:

> in accordance with the [JS Foundation IP Policy]({{ site.url }}{{ site.baseurl }}/pdf/ip-policy.pdf).

shellscape:
@Tarrask Please click the link on the comment above your last. That will take you to the site to sign the CLA.

shellscape:
@SpaceK33z could you weigh in on this one? I think it makes good sense and fixes what was probably an oversight, but I can't speak to the original intention here.

SpaceK33z:
This was intentional, but arguably it can be inconvenient. The reason that I used `options.log` for webpack warning/errors, was that `options.warn` and `options.error` were intended for warnings/errors from webpack-dev-middleware.

It makes sense to use `options.warn` and `options.error` for webpack warnings/errors, so that's fine, but it doesn't make sense to use `options.error` for the message "webpack: Failed to compile." and `options.warn` for "webpack: Compiled with warnings." This is an informational message and not a warning/error itself. If you hide these, you can't see from the terminal whether your bundle is ready.

SpaceK33z:
Note that you've changed some files from 0664 to 0775 permissions, could you fix that?

shellscape:
@Tarrask ping! please see the comments above

shellscape:
@SpaceK33z I think I should merge this and then modify with a subsequent commit to resolve the issues mentioned. There's value-add in this PR; I'd like not to hold it up further due to vanishing author 😆

shellscape:
Will be following this up with a commit to resolve the requests made.

PR #191, opened 2017-05-04
set Access-Control-Allow-Origin to * to support XHR

elevenBeans:
*I started my back-end server on localhost:5000 and front-end server on 8088.*
*When I used this middleware for HMR, I ocurred with a problem below:*

Considering that it's quite common to use 2 different servers when developing a webapp, CORS should be supported right?

*I added `Access-Control-Allow-Origin` header and set it to \**

Thanks.
BR

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -61,6 +61,7 @@ module.exports = function(compiler, options) {
 			// server content
 			var content = context.fs.readFileSync(filename);
 			content = shared.handleRangeHeaders(content, req, res);
+			res.setHeader("Access-Control-Allow-Origin", "*"); // To support XHR, etc.
 			res.setHeader("Content-Type", mime.lookup(filename) + "; charset=UTF-8");
 			res.setHeader("Content-Length", content.length);
 			if(context.options.headers) {


shellscape:
Thanks for taking the time to open a PR!

This should already be available to everyone via the `headers` options property as seen here: https://github.com/webpack/webpack-dev-middleware#usage

```js
...
headers: { "Access-Control-Allow-Origin": "*" },
...
```

Please try that and see if it works for you.

elevenBeans:
Thanks for your reply ~
But it still not works for me  : (

my server code is:
```js
var webpackMiddleware = require("webpack-dev-middleware");
var webpackConfig = require("./webpack.config");

app.use(webpackMiddleware(webpack(webpackConfig), {
	headers: { "Access-Control-Allow-Origin": "*"},
	// custom headers
}));
```
my webpack config is: [here](https://github.com/elevenBeans/WeVoting/blob/master/webpack.config.js)

shellscape:
Hm it curious that your addition at [line 64](https://github.com/webpack/webpack-dev-middleware/pull/191/files#diff-27e39f85634d9276821c67a9bbf764b4R64) would work, but setting headers in the options wouldn't, since [line 69](https://github.com/webpack/webpack-dev-middleware/pull/191/files#diff-27e39f85634d9276821c67a9bbf764b4R69) does the exact same thing. If it's not working, then that should be investigated. And I'd say that this PR should address that bug rather than adding a nearly identical line (64) to the file a few lines up from the same functionality (at line 69).

elevenBeans:
This is the problem:

Line 69 did not execute because ```context.options.headers``` is ```undefind```.
This is the ```context.options``` I recived on my server (with no 'headers').

``` json
{ "hot": true,
  "inline": true,
  "host": "localhost",
  "port": 8088,
  "publicPath": "http://localhost:8088/dist/",
  "outputPath": "/",
  "filename": "[name].bundle.js",
  "watchOptions": { "aggregateTimeout": 200 },
  "watchDelay": undefined,
  "contentBase": "/Users/libin/GitHub/WeVoting",
  "stats": 
   { "cached": false,
     "cachedAssets": false,
     "colors": { "level": 2, "hasBasic": true, "has256": true, "has16m": false },
     "context": "/Users/libin/GitHub/WeVoting"
   },
  "clientLogLevel": "info",
  "reporter": [Function: defaultReporter],
  "log": [Function: bound bound ],
  "warn": [Function: bound bound ],
  "error": [Function: bound bound ] 
}
```
Yesterday I used code below (in app server side):

```js
app.use(webpackMiddleware(webpack(webpackConfig), {
	headers: { "Access-Control-Allow-Origin": "*"},
	// custom headers
}));
```

Now, I directly add custom headers in my [webpack config file](https://github.com/elevenBeans/WeVoting/blob/master/webpack.config.js) to influence webpack-dev-server.

```js
  devServer:{
    hot: process.env.NODE_ENV === 'dev-HMR',
    inline: true,
    headers: { "Access-Control-Allow-Origin": "*" }
  }
```

Problem solved     : )

But,

+ Why the first method do not work (set the header correctly)?

+ What is the "app" refer to? Not denotes a server application ?

shellscape:
Hm I'm really not sure why it's not working on your end. [This test](https://github.com/webpack/webpack-dev-middleware/blob/23a75095bda747d24f8b902a8114dc8034303871/test/Server.test.js#L125) uses the option how it's meant to be used and it works well.

shellscape:
> What is the "app" refer to? Not denotes a server application ?

I missed this one. `app` is an instance of express, koa, etc.

PR #192, opened 2017-05-25
Parse pathname from req rather than use req.url, fix #155.

### LINKED ISSUE #155: Issue with req.url and express mounted sub-apps
Hey there,
I had some trouble getting this to work with a sub-app using express 4.x. `req.url` outputs the url relative to the sub-app and not the root app, which seems odd. As a result, this fails to match the url and returns false:
```javascript
// middleware.js, line 205
var filename = getFilenameFromUrl(options.publicPath, compiler.outputPath, req.url);
```
For example, a sub-app mounted to `/cms` with route `/static` will show `/static` instead of `/cms/static`.

This seemed to fix it:
```javascript
// middleware.js, line 205
var filename = getFilenameFromUrl(options.publicPath, compiler.outputPath, req.originalUrl);
```

sighingnow:

A bugfix, address #155.

No need for more tests for my changes.

**Summary**

In this pull request, I replace `req.url` with `parseUrl.original(req).pathname` to get correct path name of directories in HTTP request, so that we could response with the correct default files.

For example, we have the following directory structure

~~~
-- index.html
-- app
 | -- index.html
~~~

and the `options.index` is `index.html`.  When we request `/` we would get `/index.html`. However when we request `/app`, `req.url` would be `/` rather than `/app` and we would get `/index.html` rather than `/app/index.html`.

The pull request fix the issue described above.

This PR won't introduce a breaking change.

**Other information**

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -3,11 +3,10 @@
  Author Tobias Koppers @sokra
  */
 var mime = require("mime");
+var parseUrl = require("parseurl");
 var getFilenameFromUrl = require("./lib/GetFilenameFromUrl");
 var Shared = require("./lib/Shared");
 var pathJoin = require("./lib/PathJoin");
-
-var parseUrl = require('parseurl');
 
 // constructor for the middleware
 module.exports = function(compiler, options) {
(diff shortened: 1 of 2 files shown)


sighingnow:
Ping @shellscape I have finished my pull request summary as requested.

shellscape:
Thank you for the update but we still need a test added for the change.

shellscape:
@sighingnow if you get back to this and update your PR with a test, we'll be happen to reopen this one.

PR #194, opened 2017-06-01
Add time to log message

alindesign:
Small Feature 😄 

Isn't needed.

**Summary**
On my Windows system sometime file watcher don't work properly, so i want to know when was last build. So i just added time on default Reporter function.

No!


alindesign:
Ok so, i have modified tests, will work with this, about option, i can add it or is present?

PR #196, opened 2017-06-11
Make serverSideRendering example work consistently

GeorgeTaveras1231:

Documentation

N/A

**Summary**

I was following the examples and realized that it only works if you have more than one `main` bundle (such as js and on css bundle). Adding this one-liner makes it work with or without the extra `main` bundle. This small change also works when adding other named bundles (which would more than likely be single files - **non array**), and supports turning on source maps (which would make the named assets value into arrays).

Additionally, the interpolated values were returning arrays and were not being properly formatted. Therefore, I added a `.join('\n')` call at the end.

No


shellscape:
Looks good, thanks for the PR

PR #206, opened 2017-07-17
Support koa2 style middleware

iyuq:
feature

yes

**Summary**

yes

will support in koa2 framework.
```
const middleware = require("../middleware").webpackKoaMiddleware;
const Koa = require("koa");
const webpack = require("webpack");

app = new Koa();
app.use(middleware(compiler, {
  stats: "errors-only",
  quiet: true,
  publicPath: "/public/",
}));
```
**Other information**

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -8,7 +8,7 @@ var Shared = require("./lib/Shared");
 var pathJoin = require("./lib/PathJoin");
 
 // constructor for the middleware
-module.exports = function(compiler, options) {
+var exports = function(compiler, options) {
 
 	var context = {
 		state: false,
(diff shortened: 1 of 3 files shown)


iyuq:
koa2 did not support node.js < 6, so the 4.x test is failed.

shellscape:
Thanks for taking the time to put together a PR! :beer: Unfortunately I'm not sure this is the right addition for the module. 26f0b4028b15cb0e12916e1e34eac95a97ed7b8b introduced support for Koa2-like patterns and https://github.com/shellscape/koa-webpack exists to facilitate using this module with Koa2. I really don't want to see this module trying to support every server pattern there is as part of it's core functionality. It was a slippery choice to bake in express support, but made sense as it's still the most widely used server package. If koa-webpack is missing features, that's probably the best place for this work. Or, alternatively you could always create a new module. 

Of course, I could have missed the meat and potatoes of this PR, but there's no discussion to precede it. If that's the case, please let me know how I'm mistaken.

iyuq:
i know that i can create a new module to achieve this, but after that I must maintain that after the lib of this updated. If this get merged, people perefer to use Koa will be able to get things done by the wabpack organization module.
@SpaceK33z what do you think?

shellscape:
@iyuq while @SpaceK33z is the original author and who we all consider to own this module, he has passed maintenance onto several other folks due to a lack of time. (that's why some of us have a collaborator badge on our comments).

Have you looked into koa-webpack? I think that module will suit your needs, and is the proper place for additional features supporting Koa2. If I'm missing something vital, please do point it out and I'll reopen this PR.

PR #207, opened 2017-07-19
remove unnecessary if else

iyuq:

refactoring

no extra tests needed.

**Summary**

since the `res.end` method always exists, we can just use it directly.

no

**Other information**

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -78,8 +78,7 @@ module.exports = function(compiler, options) {
 				}
 				// Express automatically sets the statusCode to 200, but not all servers do (Koa).
 				res.statusCode = res.statusCode || 200;
-				if(res.send) res.send(content);
-				else res.end(content);
+				res.end(content);
 				resolve();
 			}
 		});


shellscape:
much appreciate you taking the time to open a PR, but I'm afraid I can't approve this one. that logical block isn't checking if `end` exists, it's checking for `send`, and there was a specific reason for calling `send` if it exists: eb8c505f30ede23c511f88db3597fc959bbfcac2.

iyuq:
why this is necessary? I read the related pr and find out that the commit add this if else for some version will get a res.send is not a function. Here I use res.end for all case, res.end should always exist, so what's the problem?
If I do mistake, please give me more detail which part is wrong.

shellscape:
we don't want to `end` there. we want to `send` and `end` only if the condition exists that `send` is not present.

iyuq:
In what case end here is wrong？What i know from the code is the file of the request url find successfully. if we do not need to add more content to the res, we should end it. if  we cannot end this for some case, i think should add a test case for this.

PR #209, opened 2017-07-24
fix sinon desperation

iyuq:
fix desperation method.

**Summary**
`sandbox.stub(obj, 'meth', val)` is deprecated and will be removed from the public API in a future version of sinon.
 Use `sandbox.stub(obj, 'meth').callsFake(fn)` instead in order to stub a function.
 Use `sandbox.stub(obj, 'meth').value(fn)` instead in order to stub a non-function value.

**Other information**


shellscape:
thank you for updating the tests!

PR #212, opened 2017-08-08
fix koa status code default set to 404

iyuq:
bugfix

yes

**Summary**
koa's [response](https://github.com/koajs/koa/blob/master/lib/response.js#L148) code has the following:
```
set status(code) {
    ...

    // set the status
    if (!this._explicitStatus) this.status = 200;

    ...
  }
```

We can see that if the `this._explicitStatus` is not `true`, koa will set the status to 200 when the set body is called. 

But if you do `res.statusCode = res.statusCode || 200`, it will cause the this._explicitStatus to become true, because koa default set status to 404, this is equal to `res.statusCode = 404`, so this will cause a problem.

**Other information**

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -76,8 +76,6 @@ module.exports = function(compiler, options) {
 						res.setHeader(name, context.options.headers[name]);
 					}
 				}
-				// Express automatically sets the statusCode to 200, but not all servers do (Koa).
-				res.statusCode = res.statusCode || 200;
 				if(res.send) res.send(content);
 				else res.end(content);
 				resolve();
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "webpack": "^1.0.0 || ^2.0.0 || ^3.0.0"
   },
   "dependencies": {
+    "koa": "^2.3.0",
     "memory-fs": "~0.4.1",
     "mime": "^1.3.4",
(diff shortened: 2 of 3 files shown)


shellscape:
> I really want to make sure you know how much your time and contributions are appreciated - they really really are.

The line you propose to remove was put in there for platforms like Koa, but not exclusive to Koa, which needed to have a default status set. 

We really try to be platform agnostic with the code in this module. Making changes which benefit "express-style" platforms, but which don't target one specifically, are OK. But if we start to make changes for the nuances in every possible consumer of the module, this thing is going to get pretty hairy pretty quickly. We then leave it up to consumers of the middleware to adjust for nuances. Not attempting to promote one module over another; [koa-wepback](https://github.com/shellscape/koa-webpack) and [koa-webpack-middleware](https://github.com/leecade/koa-webpack-middleware) are good examples of that. I have some obvious bias towards the first :smile: but I would highly recommend contributing nuanced changes specifically for Koa to one of those modules.

Closing this one, but commenting will remain open. And as always we can revisit if there is solid backing to do so.

iyuq:
if you do need a 200 status set, I think the following is more proper.
`res.statusCode === undefined && res.statusCode = 200;`, 
this will excute the set code when statusCode is undefined. This will not cause a problem to koa either.

PR #216, opened 2017-09-05
Allow override of mimetype detection with lookupMime option

joearasin:

Adds a lookupMime option so users can add custom mime type handling.

Yes.

**Summary**

The default behavior of inferring mime-types from the file extension may not be sufficient for all needs. For example, solving https://github.com/callstack-io/haul/issues/228 requires a messy workaround.

No. Things should work fine for existing users (assuming tests are fine).

**Other information**

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -69,7 +69,11 @@ module.exports = function(compiler, options) {
 				// server content
 				var content = context.fs.readFileSync(filename);
 				content = shared.handleRangeHeaders(content, req, res);
-				res.setHeader("Content-Type", mime.lookup(filename) + "; charset=UTF-8");
+				if(context.options.lookupMime) {
+					res.setHeader("Content-Type", context.options.lookupMime(filename));
+				} else {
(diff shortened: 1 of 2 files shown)


shellscape:
@joearasin thanks for taking the time to put together a PR!

I'm curious how this is different from the current `mimeTypes` option that webpack-dev-middleware supports. You can see how that option behaves within WDM here - https://github.com/webpack/webpack-dev-middleware/blob/4203bdf74ee3ce9c99ef9d1c00ff2ceedef009d5/lib/Shared.js#L35 - which utilizes `mime.define` as defined here - https://github.com/broofa/node-mime#mimedefine.

joearasin:
Ahh -- it appears not to be. I failed at reading documentation/code. We can probably punt on this -- let me give the original problem a spin using this and I'll get back to you.

joearasin:
This is definitely unnecessary.

shellscape:
👍 glad you were able to use that option!

Issue #229, opened 2017-11-17
WebAssembly files should be served with Content-Type=application/wasm


rhmoller:
* Operating System: Ubuntu 17.10
* Node Version: 8.4.0
* NPM Version: 5.4.0
* webpack version: next branch
* webpack-dev-middleware Version: 1.12.0

- [x] This is a **feature** request
- [ ] This is a **bug**

### Code

I have created a project for testing the new webassembly support in the next branch of webpack.

https://github.com/rhmoller/webpack-webassembly-tester

### Expected Behavior

The webpack-dev-server should serve the .wasm file with a Content-Type header = "application/wasm"

### Actual Behavior

The .wasm file is served with "application/octet-stream; charset=UTF-8"

It has the wrong mime-type and the charset specification should not be there.

### For Bugs; How can we reproduce the behavior?

Instructions are in https://github.com/rhmoller/webpack-webassembly-tester

### For Features; What is the motivation and/or use-case for the feature?

To be able to use webpacks wasm support when serving the project with webpack-dev-server

shellscape:
shouldn't this feature be asked [in the mime project](https://github.com/broofa/node-mime), considering https://github.com/webpack/webpack-dev-middleware/blob/master/middleware.js#L72?

rhmoller:
yes the mime project should add support for .wasm. I will report that.

but the middleware also adds a charset to the MIME type which does not work

https://github.com/webpack/webpack-dev-middleware/pull/230/files#diff-27e39f85634d9276821c67a9bbf764b4R72

rhmoller:
Support for the wasm content type is now available in mime 1.5.0. I have updated the pull request to use this new version and removed the temporary workaround.

shellscape:
Closed by #230

rhmoller:
Thank you. It works now when working from source. It would be great if you could publish a new version to npm.

David-OConnor:
I'm still having trouble with this. Using Create-React-App's Typescript version. When I run npm ls mime:

```
+-- mime@2.3.1
`-- react-scripts@1.1.4
  +-- url-loader@0.6.2
  | `-- mime@1.6.0
  `-- webpack-dev-server@2.9.4
    +-- express@4.16.3
    | `-- send@0.16.2
    |   `-- mime@1.4.1
    `-- webpack-dev-middleware@1.12.2
      `-- mime@1.6.0

PS C:\Users\david\code\dimensions\frontend> webpack --version
3.5.5


[...]

PR #230, opened 2017-11-17
serve .wasm files as application/wasm (without charset)

rhmoller:

Adds support for correctly serving .wasm files when using webpack/next WebAssembly support

yes

**Summary**

Changes middleware so it does not add charset to application/wasm MIME type, which solves #229

No

### DIFF
--- a/lib/Shared.js
+++ b/lib/Shared.js
@@ -32,6 +32,10 @@ module.exports = function Shared(context) {
 					options.filename = new RegExp("^[\/]{0,1}" + str + "$");
 				}
 			}
+			// add MIME type for WebAssembly
+			mime.define({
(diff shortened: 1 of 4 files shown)


lib/Shared.js
```diff
@@ -32,6 +32,10 @@ module.exports = function Shared(context) {
 					options.filename = new RegExp("^[\/]{0,1}" + str + "$");
 				}
 			}
+			// add MIME type for WebAssembly
```
rhmoller: This should ideally be handled by the mime module. I will report that.

shellscape: yeah I'd rather wait on upstream for that to land, implement here as a last resort

shellscape:
👍

PR #238, opened 2017-12-18
fixed name of logging lib and added link

donaldpipowitch:
Looking at the `package.json` it seems that the correct logging lib is [`loglevel`](https://www.npmjs.com/package/loglevel). This can be important, because there is actually a package called [`log-level`](https://www.npmjs.com/package/log-level) as well 😁

### DIFF
--- a/breaking-changes.md
+++ b/breaking-changes.md
@@ -7,7 +7,7 @@ that date.
 
 ## Informative Changes
 
-- logging is now handled by `log-level` and follows the same patterns as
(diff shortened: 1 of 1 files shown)


shellscape:
@donaldpipowitch thanks for the PR. please note for the future that PR/Issue templates aren't optional. we'll merge this one, but the next one will bite 😄

donaldpipowitch:
Alright, no problem ☺️ everyone handles this a bit differently in the case of non-code contributions ☺️ No problem. Will keep the template next time 👍

PR #250, opened 2018-01-23
prevent multiple identical compilations via watch startTime

shellscape:

An inconvenience/bug fix and introduction of a new option.

n/a

**Summary**

This mimics the fix found in [ff0000-ad-tech/wp-plugin-watch-offset](ff0000-ad-tech/wp-plugin-watch-offset) which was the result of a long-standing bug in `watchpack` in [webpack/watchpack/issues/25](https://github.com/webpack/watchpack/issues/25). It resolves the endless - or random multiple - compile issue by advancing the time for watching and stats startTime. This is introduced through the `watchOffset` option, which can be set to `0` to disable prevention and restore original behavior. 

I do not believe so. It's unlikely.

**Other information**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,6 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[.md]
+[*.md]
 insert_final_newline = false
 trim_trailing_whitespace = false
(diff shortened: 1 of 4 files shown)


shellscape:
It's been some weeks since this was opened and [webpack/watchpack/issues/25](webpack/watchpack/issues/25) is still an ongoing issue with no resolution. This has been merged as a stop-gap until watchpack is updated to handle the scenario. Once that issue is resolved, this change will be reverted.

PR #253, opened 2018-02-02
Allow sniffing of mime types based on content.

deepakjois:
Feature

Yes.

I added a single static file to the fixtures, but I need some help in hooking
it up with the server tests. I hope the current way I have done it, is okay.

**Summary**
If the original file being served does not have an extension, MIME
type detection based on file name fails and the MIME type defaults
to `null`.

I originally encountered this problem in Gatsby, the static site generator. I was trying
to generate HTML files with no extension and trying to access it (like `http://localhost:8080/Hello`)
and it would get served with the `application/octet-stream` MIME type because it had no extension.

This change adds code to check if a filename has no extension and perform MIME type sniffing
using the [detect-content-type] NPM module (which I created explicitly for this purpose).

[detect-content-type]: https://www.npmjs.com/package/detect-content-type

Not that I can see, but if it is preferable I can put this functionality behind an option so that it can be turned on and off.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const mime = require('mime');
+const detectContentType = require('detect-content-type');
 const urlJoin = require('url-join');
 const { getFilenameFromUrl, handleRangeHeaders, handleRequest, ready } = require('./util');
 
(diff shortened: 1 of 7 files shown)


deepakjois:
A small correction. Gatsby doesn’t use webpack-dev-middleware to load any HTML files. So this pull request does not serve its original intent. It might still be a useful feature.

shellscape:
@deepakjois thanks for taking the time to investigate the issue you were having, and for proposing a solution and putting the time into your solution. 

unfortunately we have to pass on this addition. it's a pretty rare case to start with, and only a fraction of actors are going to be generating html files for a bundle that are without extension. given that this module is only intended to be used for a development environment, it wouldn't be unreasonable to suggest that your development mode webpack config add an extension for testing. if that's not favorable, you could always go the `directory/index.html` route. if that's not favorable either, you could always keep the html file outside of the bundle for development and handle the mime sniffing upstream in middleware before `webpack-dev-middleware`.

PR #257, opened 2018-02-09
Fix getPaths for publicPaths that are URLs (#255)

### LINKED ISSUE #255: getPaths does not work with MultiCompilers that have URLs for publicPath
* Operating System: MacOS 10.12.6
* Node Version: 7.1.0
* NPM Version: 5.6.0
* webpack version: 3.10.0
* webpack-dev-middleware Version: 2.0.4

- [X] This is a **bug**

### Code

```js
const webpack = require('webpack');
const middleware = require('webpack-dev-middleware');
const express = require('express');

const app = express();
const compiler = webpack([{
    ...options for first compiler,
    output: {
        filename: "bundle.js",
        publicPath: "http://some.host:3000/dist/"
    },
}, {
    ...options for second compiler,
    output: {
        filename: "bundle.js",
        publicPath: "http://some.host:3000/dist2/"
    },
}]);

app.use(
  middleware(compiler, {
      publicPath: "http://some.host:3000/",
  })
);

app.listen(3000, () => console.log('Example app listening on port 3000!'));
```

### Expected Behavior

We should be able to fetch `http://some.host:3000/dist/bundle.js` and `http://some.host:3000/dist2/bundle.js`.

### Actual Behavior

In `getPaths` in `utils.js`, it checks for `url.indexOf(compilerPublicPath) === 0` which won't be true if `compilerPublicPath` has a schema/host since `url` will always start with `/`.

Try to use a `MultiCompiler` where each of the compilers has a `output.publicPath` that contains a schema/host.  A proposed solution would be to parse `compilerPublicPath` in these cases and just use the `pathname` along the lines of https://github.com/mwhansen/webpack-dev-middleware/commit/01b5eb94e5d8103c77403cd36c4e737184b12b35

mwhansen:

Fixes #255 by using only the path part of `compilerPublicPath` for checking against `url`.

Tests were added which correspond to existing tests for the same functionality.

**Summary**

This fixes an issue I was seeing with `webpack-dev-server` where it would not select the correct compiler from the list to serve the bundle.

I don't think this should introduce a breaking change since the use case this addresses currently does not work.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -24,12 +24,11 @@ function getPaths(publicPath, compiler, url) {
     && compilers[i].options.output
     && compilers[i].options.output.publicPath;
 
-      // handle the case where compilerPublicPath is a URL with hostname
       if (compilerPublicPath.startsWith('/')) {
         compilerPublicPathBase = compilerPublicPath;
       } else {
-        const urlObject = parse(compilerPublicPath);
-        compilerPublicPathBase = urlObject.pathname;
+        // handle the case where compilerPublicPath is a URL with hostname
+        compilerPublicPathBase = parse(compilerPublicPath).pathname;
       }
 
       // check the url vs the path part of the compilerPublicPath
--- a/lib/util.js
+++ b/lib/util.js
@@ -24,7 +24,7 @@ function getPaths(publicPath, compiler, url) {
     && compilers[i].options.output
     && compilers[i].options.output.publicPath;
 
-      if (compilerPublicPath.startsWith('/')) {
+      if (compilerPublicPath.indexOf('/') === 0) {
         compilerPublicPathBase = compilerPublicPath;
       } else {
         // handle the case where compilerPublicPath is a URL with hostname
--- a/lib/util.js
+++ b/lib/util.js
@@ -15,11 +15,25 @@ function getPaths(publicPath, compiler, url) {
   const compilers = compiler && compiler.compilers;
   if (Array.isArray(compilers)) {
     let compilerPublicPath;
+
+    // the path portion of compilerPublicPath
+    let compilerPublicPathBase;
+
     for (let i = 0; i < compilers.length; i++) {
       compilerPublicPath = compilers[i].options
     && compilers[i].options.output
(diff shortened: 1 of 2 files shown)


shellscape:
@mwhansen thanks for the PR! please do sign the CLA

lib/util.js
```diff
@@ -15,11 +15,25 @@ function getPaths(publicPath, compiler, url) {
...
       compilerPublicPath = compilers[i].options
     && compilers[i].options.output
     && compilers[i].options.output.publicPath;
-      if (url.indexOf(compilerPublicPath) === 0) {
+
+      // handle the case where compilerPublicPath is a URL with hostname
+      if (compilerPublicPath.startsWith('/')) {
+        compilerPublicPathBase = compilerPublicPath;
+      } else {
+        const urlObject = parse(compilerPublicPath);
```
ev1stensberg: this could be stripped (optional) 

Used once, so: `compilerPublicPathBase = parse(compilerPublicPath).pathname;` and rather add a comment above?

shellscape:
@mwhansen thanks for signing the CLA 🍺  please allow us some time for collaborators to review.

lib/util.js
```diff
@@ -15,11 +15,24 @@ function getPaths(publicPath, compiler, url) {
...
+    // the path portion of compilerPublicPath
+    let compilerPublicPathBase;
+
     for (let i = 0; i < compilers.length; i++) {
       compilerPublicPath = compilers[i].options
     && compilers[i].options.output
     && compilers[i].options.output.publicPath;
-      if (url.indexOf(compilerPublicPath) === 0) {
+
+      if (compilerPublicPath.startsWith('/')) {
```
shellscape: `startsWith` is convenient and semantic, but it's not as performant as `indexOf`. see: https://jsperf.com/startswith-vs-indexof-foo/1. I normally wouldn't nitpick on this, but being middleware we should take advantage of minor gains like this, and also be consistent.

mwhansen: Done

shellscape:
@mwhansen solid first contribution. thank you 🍻

Issue #263, opened 2018-02-25
Release v1.13 with webpack v4 listed as peedDependency


mikegreiling:
* Operating System: macOS 10.13.3
* Node Version: v9.5.0
* NPM Version: v5.6.0
* webpack version: 4.0.0
* webpack-dev-middleware Version: 1.12.1

- [x] This is a **feature** request
- [ ] This is a **bug**

### Code

The latest 1.x release (v1.12.2) will raise warnings when installed alongside webpack v4

https://github.com/webpack/webpack-dev-middleware/blob/v1.12.2/package.json#L7

```
  "peerDependencies": {
    "webpack": "^1.0.0 || ^2.0.0 || ^3.0.0"
  },
```

### Expected Behavior

Since webpack-dev-middleware v1.12.2 appears to be compatible with webpack v4.0.0, and since several webpack plugins still have not been updated to use webpack-dev-middleware v2.x, I think it would be beneficial to release **v1.13.0** which adds `^4.0.0` to the `peerDependencies`.

### Actual Behavior

When a project updates to webpack v4 and another dependency relies on webpack-dev-middleware ^1.12.0 as a sub-dependency, warnings will be thrown by most package managers and `yarn check` will fail with an error like so:

```
error "karma-webpack#webpack-dev-middleware#webpack@^1.0.0 || ^2.0.0 || ^3.0.0" doesn't satisfy found match of "webpack@4.0.0"
```

### For Bugs; How can we reproduce the behavior?

N/A

### For Features; What is the motivation and/or use-case for the feature?

I realize the best long-term goal is to nudge webpack libraries to update their webpack-dev-middleware dependencies, but for now, releasing a minor version bump for those stuck with v1.x would be a big benefit.

In my company's workflow CI will not pass if `yarn check` throws errors, so our peerDependencies have to match.  Since `karma-webpack` is still using v1.12.x, it is holding us back from upgrading to webpack v4 🙁

mikegreiling:
Related discussion in the karma-webpack library:
https://github.com/webpack-contrib/karma-webpack/pull/301#issuecomment-368336255

mikegreiling:
I'd open a PR for this, but it doesn't look like there is a v1.x branch I can target.

The process would be simple though:

1. create a branch off of tag `v1.12.2` called something like `v1-stable`
2. bump the version to `v1.13.0`
3. add `|| ^4.0.0` to the webpack `peerDependency`
4. tag and publish the new version 😄

[...]

PR #251, opened 2018-01-29
Allow webpack v4 and v4-beta

lencioni:

PR #265, opened 2018-02-26
Update webpack peerDependency in v1-stable branch to v4.x

mikegreiling:

This adds webpack v4 as a devDependency, updates the file-loader dependency to fix tests, updates the config schema used in the server-tests, and adds webpack ^4.0.0 to supported `peerDependencies`.

No, but I fixed broken tests resulting from this change.

**Summary**

This pull request aims to add webpack v4 support to the 1.x versions of webpack-dev-middleware for projects which have not yet updated to 2.x.

See issue: #263 

No.

**Other information**

I have updated the webpack `devDependency` and the `peerDependency` config, and ensured that all tests pass.  I have not, however, changed the version number in package.json for a v1.13 release.

Closes: #263 
/cc @shellscape

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
   - "6"
   - "7"
   - "8"
--- a/test/Server.test.js
+++ b/test/Server.test.js
@@ -45,7 +45,7 @@ describe("Server", function() {
 		it("GET request to bundle file", function(done) {
 			request(app).get("/public/bundle.js")
 				.expect("Content-Type", "application/javascript; charset=UTF-8")
-				.expect("Content-Length", "2823")
+				.expect("Content-Length", "3435")
 				.expect(200, /console\.log\("Hey\."\)/, done);
 		});
 
@@ -138,7 +138,7 @@ describe("Server", function() {
 
 		it("GET request to bundle file", function(done) {
 			request(app).get("/bundle.js")
-				.expect("Content-Length", "2823")
+				.expect("Content-Length", "3435")
 				.expect(200, /console\.log\("Hey\."\)/, done);
 		});
 	});
--- a/test/fixtures/server-test/webpack.config.js
+++ b/test/fixtures/server-test/webpack.config.js
@@ -1,4 +1,6 @@
 module.exports = {
+	mode: "development",
+	devtool: false,
 	context: __dirname,
 	entry: "./foo.js",
 	output: {
@@ -6,7 +8,7 @@ module.exports = {
 		path: "/"
 	},
 	module: {
-		loaders: [
+		rules: [
 			{
 				test: /\.(svg|html)$/,
 				loader: "file-loader",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Tobias Koppers @sokra",
   "description": "Offers a dev middleware for webpack, which arguments a live bundle to a directory",
   "peerDependencies": {
-    "webpack": "^1.0.0 || ^2.0.0 || ^3.0.0"
+    "webpack": "^1.0.0 || ^2.0.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {
     "memory-fs": "~0.4.1",
@@ -17,14 +17,14 @@
     "codecov.io": "^0.1.6",
     "eslint": "^4.0.0",
     "express": "^4.14.0",
-    "file-loader": "^0.11.2",
+    "file-loader": "^1.1.9",
     "istanbul": "^0.4.5",
     "mocha": "^3.0.2",
     "mocha-sinon": "^2.0.0",
     "should": "^11.1.0",
     "sinon": "^2.3.8",
     "supertest": "^3.0.0",
-    "webpack": "^3.0.0"
+    "webpack": "^4.0.0"
   },
   "license": "MIT",
   "engines": {


mikegreiling:
travis is failing here because node 4 is no longer supported by webpack.  however I don't think this is an issue because those using webpack-dev-middleware can still use older versions of webpack that support older node versions.  this isn't actually a breaking change, just an internally breaking change since the tests now target webpack v4.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "4"
```
shellscape: unfortunately we can't remove the Node v4 tests for v1.x of the module.

test/Server.test.js
```diff
@@ -45,7 +45,7 @@ describe("Server", function() {
 		it("GET request to bundle file", function(done) {
 			request(app).get("/public/bundle.js")
 				.expect("Content-Type", "application/javascript; charset=UTF-8")
-				.expect("Content-Length", "2823")
```
shellscape: is this change specific to webpack@4?

test/Server.test.js
```diff
@@ -138,7 +138,7 @@ describe("Server", function() {
 
 		it("GET request to bundle file", function(done) {
 			request(app).get("/bundle.js")
-				.expect("Content-Length", "2823")
```
shellscape: is this change specific to webpack@4?

test/fixtures/server-test/webpack.config.js
```diff
@@ -1,12 +1,14 @@
...
+	devtool: false,
 	context: __dirname,
 	entry: "./foo.js",
 	output: {
 		filename: "bundle.js",
 		path: "/"
 	},
 	module: {
-		loaders: [
+		rules: [
```
shellscape: is this change specific to webpack@4?

.travis.yml
mikegreiling: Webpack v4 does not support node 4.  We could continue to keep the `devDependency` at webpack v3.x for the sake of these tests though.

All I really wanted to do is update the `peerDependency` to allow for v4.x.

test/Server.test.js
mikegreiling: Yes, the `Content-Length` changed because webpack v4’s runtime code has changed between versions and now takes up a bit more space.

test/Server.test.js
mikegreiling: same as above 👆🏻

test/fixtures/server-test/webpack.config.js
mikegreiling: Actually this schema name change started in webpack v2 (`loaders` was deprecated in favor of `rules`), but it wasn’t removed until v4.

shellscape: OK this change can stick around

mikegreiling:
@shellscape here’s my thoughts.  Unfortunately while webpack-dev-middleware appears to be compatible with webpack v2 through v4, we can only run our tests against one version of webpack at a time because we can only specify a single version in our `devDependencies`.  As I presume most people using the legacy branch will be using webpack v2 or v3, perhaps we ought to keep webpack v3 as the test target.

However there is no reason we cannot add v4.x as a valid `peerDependency`.  Webpack v4 clearly does work with webpack-dev-middleware v1.12 as this PR demonstrates.  All of the changes I made were to the test fixtures, not to the code of the library itself.

Let me know how you’d like me to proceed.

.travis.yml
shellscape: Well in fairness, you can't remove a node dependency in a patch for major version that previously supported it. That's a massive semver violation, and that's why v2.x dropped support for it. Node 4 will have to stay and you'll have to find a way to not run webpack v4 tests on that node version. These are the things that bite you when trying to update an old version instead of upgrading.

mikegreiling: We aren’t actually removing support for node 4 with this change.  We just wouldn’t be running tests against it (because that is not possible with webpack v4).  If we kept webpack v3 as the `devDependency` version we could keep running tests against node v4.

Really this situation isn’t dissimilar from what you’re already practicing.  Webpack-dev-middleware technically supports webpack v1 and v2 but no tests are run for these versions... ¯\_(ツ)_/¯

shellscape:
You're going to have to do a few things here to get this pushed through:

1. All tests on Travis CI must run and succeed on both webpack v3 and webpack v4. 
2. Values which are different between webpack versions will have to be intelligently asserted in tests.
3. Tests for webpack v4 cannot run under Node v4

To get the webpack version:

```js
const webpackPackage = require('webpack/package.json');
const webpackVersion = parseInt(webpackPackage.version, 10);
```

I've never setup a complex travis config like what's needed here, but I do have several modules running against webpack v3 and v4. Since webpack@3 is the default devDep, `.travis.yml` will need to look something like this:

```yml
script:
  - npm run ci                  # run linting and testing, generate coverage
  - npm install webpack@latest  # install webpack@4
  - npm run mocha               # run mocha tests against webpack@3, without coverage

```

It's a lot, but we can't just shoehorn support in without being absolutely sure that we're not breaking a (past) major version. The one thing not covered here is why you haven't upgraded to v2 and called it a day.

mikegreiling:
I can do this if it is what you require, however as I mentioned earlier it doesn’t look like this approach was taken for support of webpack v1 and v2.  Both of these are supported as `peerDependencies` and there are no tests being run with these versions.

shellscape:
I wasn't a maintainer for the versions that covered webpack v3 an v4 😄  and no version has been as disruptive as webpack v4 to date. Releasing new major versions is how we support the new majors of the primary dependencies. Is it worth the time to do all of this, or simply upgrade to webpack-dev-middleware v2.x?

mikegreiling:
I suppose it probably would have been faster to upgrade `karma-webpack` to use v2.x, but this seemed like a simpler approach as I already knew it worked with v1 and just needed a `peerDependency` change in order to fix the warnings I was seeing.  I assumed the same was likely true for other libraries which haven’t upgraded yet so I figured fixing something once here would be less effort than fixing something in several other places.

Since I’ve already done half the work, I’ll see if I can just finish the job here and add support for v3 and v4 in the travis config.  I presume this change would also be welcome in the v2 branch since it appears you wish to continue supporting webpack v3 there as well.

mikegreiling:
I would like to note also that the current version of webpack-dev-middleware v2 you’ve released supports webpack v4 as a `peerDependency` without any tests (see #251 )...  Was that a mistake?  I’m not sure why this wasn’t worthy of the same exception.

shellscape:
I don't know how much you followed the webpack v4 alpha/beta/release cycle, but it was impossible to keep up with. Hence no formal suite of tests covering it. I'm presently going through that pain on [webpack-serve](https://github.com/webpack-contrib/webpack-serve/tree/webpack-4-support) and [webpack-hot-client](https://github.com/webpack-contrib/webpack-hot-client/pull/20/files) so I'm also not sure if that was a dig at me about dev-middleware not being up to par yet... but I'm only one human, cranking out free code in the free time that I can manage.

If I have my way, every new major will have tests running in CI to cover the last two versions of webpack. That is until we just start mandating that new major versions of companion modules only support the current version. Be glad that someone is making this push so folks a version back on webpack don't get left in the dust.

mikegreiling:
That was not meant as a dig at you personally.  I appreciate you taking the time to be thorough. 😅  I was just pointing out what appeared to me to be a slight double standard – one that you may not have been conscious of.

I’ll update this PR as soon as I can find the time, assuming you’re still ok with merging it.  I’ll update #266 as well.

shellscape:
Yessir, I'm still OK to merge it as long as we have our bases covered.

FWIW in this ecosystem (webpack) it's not so much a double standard as it is misalignment or inconsistency. Much of the time it's just simply a lack of time to do all the things needed to maintain that alignment.

shellscape:
@mikegreiling is this still a need for your team?

mikegreiling:
@shellscape I had intended to return to this but got I busy with other things.  In the mean time it looks like `karma-webpack` has updated its `webpack-dev-middleware` dependency with https://github.com/webpack-contrib/karma-webpack/pull/318

I’m going to close this now as it is no longer needed for me.  Thanks for being willing to work with me on it :+1:

PR #275, opened 2018-03-02
Don't add / to url when in Windows

### LINKED ISSUE #270: Bad bundle filename generated on Windows
* Operating System: Windows 10 x64
* Node Version: 8.4.0
* NPM Version: 5.3.0
* webpack version: 4.0.1
* webpack-dev-middleware Version: 3.0.0

- [x] This is a **bug**

### Code

webpack.config.js snippet:

```js
    output: {
        filename: 'bundle.js',
        path: path.join(__dirname, 'dist'),
        publicPath: '/js/',
        libraryTarget: 'window'
    },
```

server express snippet:

```js
        const webpack = require('webpack')
        const webpackDevMiddleware = require('webpack-dev-middleware')
        const config = require('webpack.config.js')
        const compiler = webpack(config)
        app.use(webpackDevMiddleware(compiler, {
            publicPath: config.output.publicPath
        }))
```

### Expected Behavior

webpack-dev-middleware is able to serve the generated bundles correctly on Windows.

Given an `output.path` of `c:\workspace\dist` and a filename of `bundle.js` the generated filename from `getFilenameFromUrl` should be: `c:\workspace\dist\bundle.js`

### Actual Behavior

Fails to find the generated bundle due to a malformed Windows path when joining the bundle name and the output path, as the bundle path generated is in the form: `/c://\workspace\dist/bundle.js`.

This appears to be because of code like https://github.com/webpack/webpack-dev-middleware/blob/78e98794e5fff0728c1d33657cddbcc192a6ccb7/lib/util.js#L94 that uses `url-join` instead of Node's `path` module for forming file paths, plus the bit at https://github.com/webpack/webpack-dev-middleware/blob/78e98794e5fff0728c1d33657cddbcc192a6ccb7/lib/util.js#L96 that slaps `/` at the start of a path instead of using `path.resolve`. Those malformed URIs are then passed into the filesystem which naturally fails to find the file.

Replacing that chunk of code at the end of `getFilenameFromUrl` with the simpler `return path.resolve(outputPath || '', filename)` fixes the problem for me, though I'd hardly be surprised if that broke other use cases, so the proper fix is perhaps more intricate.

I note that some other use cases in the code like https://github.com/webpack/webpack-dev-middleware/blob/78e98794e5fff0728c1d33657cddbcc192a6ccb7/lib/middleware.js#L49 are making the same mistake of using `url-join` instead of `path`, but they aren't affecting my use case so I haven't needed to touch them.

I wouldn't be surprised if the current wdm code breaks in some rather unlikely setups with POSIX paths, too. As similar as POSIX paths and URIs are, they don't have exactly identical rules and it's a mistake to treat them as interchangeable.

Run a webpack-dev-middleware example on Windows.

### For Features; What is the motivation and/or use-case for the feature?

Developing on Windows workstations.

montogeek:
Adds an aditional check for Windows platform when processing assets URL.

No

**Summary**

Fixes #270

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -9,6 +9,7 @@ const urlJoin = require('url-join');
 const DevMiddlewareError = require('./DevMiddlewareError');
 
 const HASH_REGEXP = /[0-9a-f]{10,}/;
+const isWindows = process && (process.platform === 'win32' || /^(msys|cygwin)$/.test(process.env.OSTYPE));
 
 // support for multi-compiler configuration
 // see: https://github.com/webpack/webpack-dev-server/issues/641
@@ -93,7 +94,7 @@ module.exports = {
     if (filename) {
       uri = urlJoin((outputPath || ''), filename);
 
-      if (!uri.startsWith('/')) {
+      if (!isWindows && !uri.startsWith('/')) {
         uri = `/${uri}`;
       }
     }


seanmiddleditch:
This solves the issue in my immediate use case.

I'd be concerned that no tests are added for this change to ensure that NT paths continue working in the future, but otherwise, it gets the job done. :)

montogeek:
I actually don't like this fix, a better solution would use a library that handle all this cases across different systems.

shellscape:
I spoke with @montogeek offline and he agreed to move forward with #276 instead of this PR. Thanks for the efforts in trying to debug this one!

PR #276, opened 2018-03-03
Fix Windows Paths

seanmiddleditch:

Bug fix for #270 .

Test added for the Windows path that will be generated after this fix.

Where we might expect `c:\foo\bar` we instead get a URL-ized `c://\foo/bar`. Either works with the default path translation on Windows, though *technically* it's wrong.

**Summary**

Allows using webpack-dev-middleware on Windows. Without this, the path generated is along the lines of `/c:\foo/bar` which is then translated by Windows as `c:\c:\foo\bar` under the hood, which won't resolve properly.

Not that I'm aware of. It's possible that someone using a very non-idiomatic path on a POSIX system might have problems, maybe.

**Other information**

I wouldn't call this the best possible fix, but It Werks(tm). A more correct change would run much deeper, perhaps requiring modifications to the file system abstraction in Webpack itself.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -93,7 +93,7 @@ module.exports = {
     if (filename) {
       uri = urlJoin((outputPath || ''), filename);
 
-      if (!pathabs(uri)) {
+      if (!pathabs.posix(uri) && !pathabs.win32(uri)) {
         uri = `/${uri}`;
       }
     }
--- a/lib/util.js
+++ b/lib/util.js
@@ -93,7 +93,7 @@ module.exports = {
     if (filename) {
       uri = urlJoin((outputPath || ''), filename);
 
-      if (!uri.startsWith('/')) {
+      if (!pathabs(uri)) {
         uri = `/${uri}`;
       }
     }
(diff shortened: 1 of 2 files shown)


shellscape:
Awaiting your followup on #270 to @montogeek's fork/fix (which is different than the commit I made).

> Where we might expect `c:\foo\bar` we instead get a URL-ized `c://\foo/bar`. Either works with the default path translation on Windows, though *technically* it's wrong.

That's not exactly preferable either. I'm not a huge fan of bandaids where there's a larger issue at play. Since v2.x is still viable we can afford to fix this all the way around.

And we know you're frustrated by this, but let's move forward without the needless snark 😄

seanmiddleditch:
Commented on his PR. I'd argue that this change is slightly preferable since it adds a test for NT-compatible paths that will be executed on the Travis CI test, but otherwise they're basically doing the same thing and either change gets the job done. :)

shellscape:
Hey all. I'm trying to get AppVeyor setup for this repo so we can get some Windows tests running in CI to avoid a similar situation.

shellscape:
OK AppVeyor is setup but will require another PR to kick off the builds. I'm going to merge this one for now. Battling the flu but I'm going to see about getting a patch release out today.

PR #285, opened 2018-03-21
Clarify best practice: Use `webpackConfig.output.publicPath`

strand:

The change is a documentation improvement.

No. Docs changes do not impact the implementation and thus do not require tests.

**Summary**

When installing this middleware I was surprised that `webpackDevMiddleware(webpack(webpackConfig)` didn't just work. When I reviewed the docs I realized that publicPath is required, and tried to add the publicPath to my webpack.config, but publicPath is not a root level property for this config option. Instead the publicPath is nested under the output property, and we should specify that the `output.publicPath` is the correct property to assign `publicPath` to.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,8 +151,7 @@ eg. `{ 'text/html': [ 'phtml' ] }`. Please see the documentation for
 Type: `String`  
 _Required_
 
-The public path that the middleware is bound to. _Best Practice: use the same
-`publicPath` defined in your webpack config._
+The public path that the middleware is bound to. _Best Practice: use your webpack config's `output.publicPath` as the value for the required `publicPath`. For example, if you've required your webpack config as `webpackConfig`, declare `publicPath: webpackConfig.output.publicPath` within the options you pass to the middleware._
 
 ### reporter
 


strand:
I'm waiting on confirmation from my employer that I can sign this CLA.

shellscape:
I'm not exactly sure this is necessary, seems superfluous given that there are several pages in the webpack documentation dedicated to that property. see:
- https://webpack.js.org/guides/public-path
- https://webpack.js.org/configuration/output/#output-publicpath

I'd like to know what specifically about that threw you for a loop. Were you not aware of the config property, or didn't know that it aligned with `output.publichPath`? Were you aware of the documentation available? I ask as there may be an opportunity for webpack as a whole to improve here.

strand:
@shellscape 💖 I had an existing project running express, and realized I'd not configured webpack to work with express when I discovered this issue. My expectation of README's is that I should be able to understand the high-level requirements of a piece of software to configure it to do its thing. Because a best practice was listed reading `Best Practice: use the same`publicPath` defined in your webpack config.` I worked backwards, went into my config file and tried adding a publicPath to the config file and got a warning. 

To resolve this issue my preference would be to first change the README so it is more accurate, and then remove the requirement altogether. When no publicPath argument is received, we should first attempt to infer the publicPath from the `output.publicPath`, then warn the user if the config files `output.publicPath` property is not defined.

Basically, I expect modules to not make me think, and was surprised when `webpack-dev-middleware` failed to infer a reasonable default for the `publicPath`. Whilst I can look up publicPath docs, I should not have to understand how publicPath works to use your module.

----

btw, the issue that I started on is `Cannot find module 'js'` and I still cannot find simple instructions that outline the requirements of babel/webpack/any-other-magical-middleware to allow me to compile my `js` or `jsx` file.

I think that there is definitely room to improve how webpack is documented, but a lack of documentation isn't my issue… I can find lots of folks who are hitting warnings and errors without much meaning and no indication of how to troubleshoot. So, you search the web and find that there is a stack overflow, but the question is regarding three major versions ago when the configuration syntax was different.

NOTE: I know this is a hard, hard problem. It shouldn't be this difficult for new people to troubleshoot and resolve using js/jsx templates within a project using webpack, but I'm not sure what adjustment to strategy would make this easier.

shellscape:
> Basically, I expect modules to not make me think, and was surprised when webpack-dev-middleware failed to infer a reasonable default for the publicPath. Whilst I can look up publicPath docs, I should not have to understand how publicPath works to use your module.

I think this just demonstrates a lack of understanding of the tools you're trying to use. If something points you at "your webpack config," your first stop should be the webpack docs - especially if it's referencing a specific configuration settings/property and you're unaware of what that property is/does. It's not reasonable that a module thoroughly document all of it's dependencies, in light of the user not reading the docs for the dependencies which they're willfully using, especially when the module is cursory to the dependency. 

Not everyone thinks the same way about problems, and not everyone problem solves the same way. I think you've been bit by the otherthinking-it-bug here though. Start simpler, attack the obvious, learn the tools you're using. 

I'm good with appending the original README content with a link to the `publicPath` documentation, but I'll take a hard pass on your changes to that section, as I just don't think they're necessary. If you'd like to make that change, it should read:

```md
For more information about `publicPath`, please see [the webpack documentation](https://webpack.js.org/guides/public-path).
```

strand:
Will you accept a PR which removes the requirement altogether, assumes the config's `output.publicPath` as its default, and provides a warning if the `output.publicPath` is not present?

shellscape:
Nope

strand:
> Not everyone thinks the same way about problems, and not everyone problem solves the same way.

This is very, very true. One of the ways I learn things is by breaking them, and then putting them back together. Because of this I prefer tools which provide reasonable defaults and tell me how to fix common problems encountered in configuration.

I really don't understand why this middleware requires `publicPath` when all its other options are, well, optional. Why did you decide to require this?

shellscape:
There's a lot of history to why that is the case. I don't have the time to lay that all out, but you can probably glean the reasons for that by searching the issues in this repo and webpack-dev-server, which the issue originated.

strand:
Thank you for suggesting breaking it down to a simpler problem. I was trying to trace around an existing implementation from another app, and I been going down a cargo culting blind alley.

PR #293, opened 2018-04-05
refactor: use chalk from webpack-log

shellscape:

A refactor

No need

**Summary**

There's some controversy over which kind of dependency `chalk` should be and its use. `webpack-log` now exports `chalk` as a dependency for one-off use.

No

**Other information**


shellscape:
Travis is dumb again, not using the right branch to build.

Issue #284, opened 2018-03-20
%20 in directory name breaks path validation in middleware.js


wbern:
* Operating System: Windows 10 x64
* Node Version: 8.1.2
* NPM Version: 5.0.3
* webpack version: 3.11.x
* webpack-dev-middleware Version: 2.0.2

- [ ] This is a **feature** request
- [X] This is a **bug**

### Code

Directly related to PR #201
If you have %20 in any directory name leading up to your webpack build, it will be interpreted as " ", which is incorrect.

actual path: `C:\My%20Path\wwwroot\...`
middleware.js
```
let filename = getFilenameFromUrl(context.options.publicPath, context.compiler, req.url);
```
returns: `C:\My Path\wwwroot\...`

### Expected Behavior
should return `C:\My%20Path\wwwroot`

### Actual Behavior
returns: `C:\My Path\wwwroot\...`

### For Bugs; How can we reproduce the behavior?
Just make an output path with %20 in any directory name.

### For Features; What is the motivation and/or use-case for the feature?
Our repo has %20 in it, it is hard to change it...

wbern:
Here is a work-around that may work for most people stumbling into this issue. I'm basically creating a mutable reference (alias if you will) in the memoryfilesystem object.

```
const middlewareInstance = webpackDevMiddleware(webpackCompiler, {
    publicPath: config.output.publicPath
});

middlewareInstance.waitUntilValid((/* stats*/) => {
    const recurse = (pathObj) => {
        Object.keys(pathObj).forEach(key => {
            if (typeof pathObj[key] === "object") {
                // it's a path object
                if (key.indexOf("%20") !== -1) {
                    // there are %20 signs. make aliases to support real spaces too
                    // optionally, we could keep recursing down if we know there are more dirs
                    pathObj[key.replace(/%20/g, " ")] = pathObj[key];
                } else {
                    // no %20 signs, keep recursing until we find it
                    recurse(pathObj[key]);
                }
            }
        })
    }
    recurse(middlewareInstance.context.fs.data);
});

app.use(config.output.publicPath, middlewareInstance);
```

shellscape:
Thanks for the issue 🍺 

This is kinda edge-casey, but I can see how it'd be pretty inconvenient. We'd welcome a fix PR, or we can leave it as-is with the workaround you've posted for folks.

shellscape:
@wbern I have a fix in progress, but I need to see what your webpack config looks like.

PR #303, opened 2018-04-26
Excluded outputPath from URI escaping to fix #297.

### LINKED ISSUE #297: parent directories with spaces don't seem to be supported
* Operating System: Windows 10 Home v1709 Build 16299.371
* Node Version: 9.11.1
* NPM Version: 5.6.0
* webpack Version: 4.4.0
* webpack-dev-middleware Version: 3.1.2

- [x] This is a **bug**

Duplicate issue from webpack-dev-server and webpack-server, difference is that having a space in the directory path with webpack-serve gives a `Not Found`  instead of a directory listing.

Information and problem details found (specific to webpack-dev-server, but same issue) [here](https://stackoverflow.com/questions/49763540/webpack-webpack-dev-server-react-shows-directory-listing-when-it-should-show-app)

Found this from tracking down this [Webpack-dev-server issue](https://github.com/webpack/webpack-dev-server/issues/1375) and then this [webpack-serve issue](https://github.com/webpack-contrib/webpack-serve/issues/100)

Was redirected to `webpack-dev-middleware` from `webpack-serve` issue

### Code

Not my code, but a simple starter repo I've used for testing:
https://github.com/bradtraversy/react_webpack_starter

### Expected Behavior
Bundled app runs, works on project when there are no spaces in absolute directory path.

### Actual Behavior
When there are spaces in the absolute directory path: 

- Shows `Not Found` on projects using webpack-serve
- Shows directory listing on projects using webpack-dev-server 

See github issue for more information, that issue is specific to webpack-dev-server.

For webpack-serve, see [here](https://github.com/webpack-contrib/webpack-serve/issues/100)
For webpack-dev-server, see [here](https://github.com/webpack/webpack-dev-server/issues/1375)

stephenmartindale:
This change fixes the regression tracked by #297 and introduced in https://github.com/webpack/webpack-dev-middleware/commit/08207ccf7759fdcea38566ea8848426c3f6c36bf.

To support both spaces in `outputPath` *and* URI-escape-like sequences such as `%20`, I have simply excluded `outputPath` from URI unescaping under Windows. The filename part is still unescaped.

Tests have been added - see changes for an illustration of what should now work. One test for #284 has been removed because it was invalid: %20, in a URI, should be resolved to space, always.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -89,24 +89,32 @@ module.exports = {
 
     let uri = outputPath;
 
-    if (filename) {
-      uri = urlJoin((outputPath || ''), filename);
-
-      if (!pathabs.posix(uri) && !pathabs.win32(uri)) {
-        uri = `/${uri}`;
-      }
-    }
-
-    // if no matches, use outputPath as filename
-    result = querystring.unescape(uri);
-
-    // fixes #284, on windows path spaces should resolve to %20
     /* istanbul ignore if */
     if (process.platform === 'win32') {
-      result = result.replace(/\s/g, '%20');
-    }
+      // Path Handling for Microsoft Windows
+      if (filename) {
+        uri = urlJoin((outputPath || ''), querystring.unescape(filename));
 
-    return result;
+        if (!pathabs.win32(uri)) {
+          uri = `/${uri}`;
+        }
+      }
+
+      return uri;
+    }
+    else {
+      // Path Handling for all other operating systems
+      if (filename) {
+        uri = urlJoin((outputPath || ''), filename);
+
+        if (!pathabs.posix(uri)) {
+          uri = `/${uri}`;
+        }
+      }
+
+      // if no matches, use outputPath as filename
+      return querystring.unescape(uri);
+    }
   },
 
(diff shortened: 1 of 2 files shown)


shellscape:
@stephenmartindale thanks a ton for getting a PR in for this. 🍺  Looks like you have some linting issues in CI. You can check that before pushing by running `npm run lint`.

_Please Note: It's never cool to remove a PR or Issue template. Please don't remove the template in the future._

shellscape:
@stephenmartindale it's possible that by reversing the commit you referenced, another bug was unmasked or produced. unfortunately that'll have to be resolved before we can merge this, as we can't merge a change with a failing test.

stephenmartindale:
Ok. Done.

For the record, the Windows-only tests do pass on my Windows machine. Running them on a Linux server makes little sense because the path handling is dramatically different. However, it could be done by adding an optional `platform` parameter to `getFilenameFromUrl()` (defaulting it to `process.platform` when `undefined`) but that's a breaking change to the signature of the method and outside the scope of this patch.

PR #311, opened 2018-07-02
updates example - catches the case whereby assets are an object

311289189:
Change to the SSR example in the README

tests not required

**Summary**
when following this example for SSR I added a check for the case in `normalizeAssets` whereby an object will be converted to an array. This makes sense for the purpose when your webpack entry is an object. Then you will get an array of paths.

no

**Other information**
I'm not an expert and maybe there is an edge case where this method is unsuitable, but it works for my use case and would maybe help someone else too

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,9 @@ const middleware = require('webpack-dev-middleware');
 
 // This function makes server rendering of asset references consistent with different webpack chunk/entry configurations
 function normalizeAssets(assets) {
+  if (typeof assets === 'object') {
+    return Object.values(assets)
+  }
   return Array.isArray(assets) ? assets : [assets]
 }
 
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ const middleware = require('webpack-dev-middleware');
 
 // This function makes server rendering of asset references consistent with different webpack chunk/entry configurations
 function normalizeAssets(assets) {
-  if (typeof assets === 'object') {
+  if (Object.prototype.toString.call(assets) === '[object Object]') {
     return Object.values(assets)
   }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -318,7 +318,7 @@ const middleware = require('webpack-dev-middleware');
 
 // This function makes server rendering of asset references consistent with different webpack chunk/entry configurations
 function normalizeAssets(assets) {
-  if (typeof assets === 'object') {
+  if (Object.prototype.toString.call(assets) === '[object Object]') {
```
shellscape: this is not a good change, as there are many different cases which qualify an entity as an object. you should recommend use of a module like `is-plain-obj` instead

311289189: @shellscape Yes I'm just realising this now, will change

ev1stensberg: `isObject` from lodash is okay, but it will make a big dependency. When checking object type i usually go for `typeof assets === Object` or `assets instanceof Object`

311289189: @ev1stensberg but the problem there is 
image

311289189: either of those checks would catch arrays

shellscape: yeah just go with a package to make this example simple. we don't need to bikeshed over that, we just want something simple for users to consume in the docs. if they prefer another method, I'm sure they'll use it.

311289189: 👍

evilebottnawi:
@michael-ciniawsky same problem with tests here?

michael-ciniawsky:
Yes, it's a flake (we need to investigate separately) :) The docs update likely won't have had an effect here :D 😛

Issue #310, opened 2018-05-21
Unable to write asset to disk: TypeError: chalk is not a function


Izhaki:
* Operating System: OSX 10.12.6
* Node Version: 8.4.0
* NPM Version: 6.0.1
* webpack version: 4.8.1
* webpack-dev-middleware Version: 3.1.3

- [X] This is a **bug**

This relates to https://github.com/webpack/webpack-dev-middleware/issues/291

### Code

```js
const webpack = require('webpack');
const middleware = require('webpack-dev-middleware');
const webpackConfig = require('../build/webpack.dev');
const compiler = webpack(webpackConfig);

const app = express();

app.use(middleware(compiler, {
  publicPath: '/app/scripts_amd',
  stats: {
    cached: false,
  },
  writeToDisk: true
}));
```

### Expected Behavior

Files being written to disk.

### Actual Behavior

```
TypeError: chalk is not a function
✖ ｢wdm｣: Unable to write asset to disk:
```

### How can we reproduce the behavior?
It seems you only need to install the versions described above, with the config above.

### More details.

With 3.1.3:

This seems to be the result of this PR https://github.com/webpack/webpack-dev-middleware/pull/293

[/lib/fs.js](https://github.com/webpack/webpack-dev-middleware/blob/27b6c1af28487263d30888751982b979ff267941/lib/fs.js#L7):

    const { chalk } = require('webpack-log');

[package.json](https://github.com/webpack/webpack-dev-middleware/blob/27b6c1af28487263d30888751982b979ff267941/package.json#L42)

    "webpack-log": "^1.0.1"

But [that version of webpack-log does not export chalk](https://github.com/webpack-contrib/webpack-log/blob/v1.0.1/index.js). Version [1.2.0 does](https://github.com/webpack-contrib/webpack-log/blob/v1.2.0/index.js).

### Fix:
Upgrade `webpack-log` to 1.2.0.

shellscape:
Thanks for the issue 🍺 

The semver pattern `^1.0.1` matches `1.2.0` so this is probably an issue on your end. All tests pass and a fresh install (npm 6.x should respect package-lock).

Izhaki:
Right then!

So there we have it... on a scale...

# Option 1: ~45 minutes per person.

[...]

PR #318, opened 2018-08-14
Update webpack-log dependency to 1.2.0

### LINKED ISSUE #317: Version number of webpack-log dependency is incorrect
Issue #310 rightly pointed out that the version 1.0.1 of webpack-log does not work with webpack-dev-middleware, so "^1.0.1" is an invalid semvar dependency specification for this package. "^1.0.1" indicates that version 1.0.1 should work, which is false. The resolution of "^1.0.1" to a later version of webpack-log on a fresh install is not a valid argument, for multiple reasons:

1. It disregards the semvar specification defining the meaning of the ^ syntax.
2. A fresh install of node_modules should not be required to use webpack-dev-middleware.
3. A project may have already installed version 1.0.1 (or some other version that matches "^1.0.1" yet is incompatible with webpack-dev-middleware), and therefore the correct later version (1.2.0) will not be installed.

As the reporter of #310 (somewhat snarkily) pointed out, this trouble could be avoided by you folks being good stewards of your package.json and updating the version number of webpack-log to match only the versions that are compatible with webpack-dev-middleware.

I will make it easy for you by contributing the PR momentarily.

svicalifornia:
Fixes #317.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "path-is-absolute": "^1.0.0",
     "range-parser": "^1.0.3",
     "url-join": "^4.0.0",
-    "webpack-log": "^1.0.1"
+    "webpack-log": "^1.2.0"
   },
   "devDependencies": {
     "assert": "^1.4.1",


svicalifornia:
So updating a dependency to the version mutually recommended in #310 breaks your test suite. Well, that's fun. I'll have to come back to fix this later, unless someone else can help sooner.

evilebottnawi:
@michael-ciniawsky can we fix it in this PR?

michael-ciniawsky:
I think so, but I'm not 💯 sure if the `content-length` check has an important meaning or not, from the test suit it looks like it doesn't (`app.js` is e.g not an injected script by the middleware or the like, just a test `app.js` bundle (which can change in size easily)). I would suggest to update the size to the currently recieved value for now and add a comment that this needs triage anf refactor from our side :)

evilebottnawi:
@michael-ciniawsky let's do it, feel free create separate PR for this

michael-ciniawsky:
No 😞 and I just saw you also don't have 🤢...

evilebottnawi:
@michael-ciniawsky waiting for access

PR #319, opened 2018-08-20
New option: acceptedMethods

ferdinando-ferreira:
Feature

Yes

**Summary**
This change adds a new option to the middleware: `acceptedMethods`, to allow for the middleware, optionally, to accept methods other than just GET. 

No, given that the new option is is optional and that its default value results in the same previously existing behaviour there won't be a breaking change.

**Other information**
The impact on the middleware itself is very limited, the purpose of this feature would be to allow for (let's say) webpack-serve, using the serve.devMiddleware option, to set this option and allow for the middleware to respond to methods other than only GET. Below is an excerpt of a webpack.config.js using this option along with a (soon to be proposed) similar option 

```javascript
  serve: {
  	hotClient: { host:  { client: '*', server: '0.0.0.0' } },
  	host: '0.0.0.0',
  	devMiddleware: {
  		publicPath: ProjectConfig.getOutput(target).publicPath,
  		acceptedMethods: ['GET', 'POST']
  	},
  	add: (app, middleware, options) => {
  		const historyOptions = {};
  		historyOptions.logger = console.log.bind(console);
  		historyOptions.acceptedMethods = ['GET', 'POST'];
  		app.use(convert(history(historyOptions)));
  	}
  }
```

The immediate effect of this change is that, when the action of a form inside webpack-serve tries to submit it to  the application it is passed along by the middleware (to be treated by the application) instead of simply returning a 404.


README.md
```diff
@@ -62,6 +62,13 @@ for the Object.
 
 _Note: The `publicPath` property is required, whereas all other options are optional_
 
+### acceptedMethods
```
michael-ciniawsky: `options.methods` or even `options.methods.accept(ed)` to be consistent with naming `e.g options.headers` ? (Ideally we cold group it under a `options.http` prop but that's sadly not possible atm)

ferdinando-ferreira: > Ideally we cold group it under a options.http prop but that's sadly not possible atm
Is it a limitation caused by interaction with some other module? I ask that because it is possible, with an (otherwise) inconsequential change at https://github.com/webpack/webpack-dev-middleware/blob/ebdcedadbe485eb7cc4cfcacd57af544977cd2cb/lib/middleware.js#L27 
to make it
```javascript
const acceptedMethods = (context.options.http ? context.options.http.methods : null) || ['GET'];
```

The webpack.config would then be
```javascript
serve: {
  hotClient: { host:  { client: '*', server: '0.0.0.0' } },
  host: '0.0.0.0',
  devMiddleware: {
    publicPath: ProjectConfig.getOutput(target).publicPath,
    http: {
      methods: ['GET', 'POST']
    }
  },
  add: (app, middleware, options) => {
    const historyOptions = {};
    historyOptions.logger = console.log.bind(console);
    historyOptions.acceptedMethods = ['GET', 'POST'];
    app.use(convert(history(historyOptions)));
  }
},
```
I tested it in my environment and it works, what would be the impediment to use options.http.methods?

michael-ciniawsky: 🤔 I tink this could work out and would personally be in favor of making this change. It will likely involve some modifictaions to `webpack-dev-server` but that's ok from my side

cc @evilebottnawi What do you think about adding it to `options.http` instead. Anything of  concern with this approach ?

michael-ciniawsky: It's definitely more organized and we could eventually move `options.headers` under `options.http`  (e.g `options.http.headers`) in the next major aswell then

michael-ciniawsky: Yep, `options.methods` is equally fine aswell, we can eventually group them later in the next major, but it's a separate issue. So we all agree on moving forward with `options.methods` ? @ferdinando-ferreira ?

PR #321, opened 2018-08-20
chore(package): add `release` script (`scripts`)

michael-ciniawsky:


michael-ciniawsky:
Yes, but we need to polish it a bit first, otherwiese it will e.g install `loader-utils`  and `schema-utils` and a ton of scriots not needed here :). I add it to the `ROADMAP`

PR #330, opened 2018-08-22
feat(middleware): add `generateContent` option (`options.generateContent`)

ferdinando-ferreira:
Feature

Yes

**Summary**
This change adds a new option to the middleware: `generateContent`, to allow for the middleware, optionally, to accept a function that is capable of modifying the served files on the fly. 

No, given that the new option is is optional and that its default value results in the same previously existing behaviour there won't be a breaking change.

**Other information**
The impact on the middleware itself is very limited, the purpose of this feature would be to allow for (let's say) webpack-serve, using the serve.devMiddleware option, to set this option and allow for the middleware to change files on the fly (for instance, adding a script to the html, adding a watermark to served images) or to signal to the middleware that it must not serve that file. Below is an excerpt of a webpack.config.js using this option (also featured in the tests and the documentation sample in a more simplified fashion). 

```javascript
serve: {
  devMiddleware: {
    publicPath: this.getOutput(target).publicPath,
    methods: ['GET', 'POST'],
    generateContent: (context, filename, req, res, defaultGenerator) => {
      const serialize = require('serialize-javascript');
      let content = defaultGenerator(context, filename, req, res);
      if (/\.html$/.test(filename)) {
        const search = '<!--self generated vars-->';
        let content_string = content.toString();
        if ((new RegExp(search)).test(content_string)) {
          let obj = {
            method: req.method
          };
          let replace = `<script type="text/javascript">var global_context = ${serialize(obj)};</script>`;
          let replaced = content_string.replace(search, replace);
          content = Buffer.from(replaced);
        }
      }
      return content;
    }
  }
}
```

The effect is to transform the following .html (served in a typical vue SPA)
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>plataforma</title>
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
    <meta name="mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-capable" content="yes">
  </head>
  <body>
    <div id="app"></div>
    <!--self generated vars-->
    <script type="text/javascript" src="/index.js"></script>
  </body>
</html>
```

into this

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>plataforma</title>
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
    <meta name="mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-capable" content="yes">
  </head>
  <body>
    <div id="app"></div>
    <script type="text/javascript">var global_context = {"method":"GET"};</script>
    <script type="text/javascript" src="/index.js"></script>
  </body>
</html>
```

Best regards

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webpack-dev-middleware",
-  "version": "3.1.1",
+  "version": "3.1.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {


README.md
```diff
@@ -69,6 +69,45 @@ Default: `[ 'GET' ]`
...
 This property allows a user to pass the list of HTTP request methods accepted by the server. 
 
+### generateContent
+
+Type: `Function`  
+Default: `undefined`
+
+This option accepts a `Function` value which is called before the files being served are read from the filesystem. It can be used to modify the file on the fly, skip their processing altogether or simply log their generation. For instance, the following configuration will replace the string `<!--self generated vars-->` with `<script type="text/javascript">var global_request = { method: "GET" }</script>` in every .html file requested via HTTP GET.
+
+```js
```
michael-ciniawsky: This example is definitely too complicated for description :)

lib/middleware.js
```diff
@@ -65,7 +65,12 @@ module.exports = function wrapper(context) {
         }
 
         // server content
-        let content = context.fs.readFileSync(filename);
+        const defaultGenerator = ((context, filename, req, res) => context.fs.readFileSync(filename));
```
michael-ciniawsky: Why passing `res, req`  as params ? (Use Cases)

ferdinando-ferreira: It is the signature of the function, the parameters that today are currently available to generate the content. 

It makes no difference today but, in the future, they may be required for the base middleware to perform some filtering, some operation (not foreseeable today)  and adding them back to the function would be a breaking change. Basically it is a case of the child function passing back what they received in the first place, to make it easier to chain.

ferdinando-ferreira: It is the signature of the function, the parameters that today are currently available to generate the content. 

It makes no difference today but, in the future, they may be required for the base middleware to perform some filtering, some operation (not foreseeable today)  and adding them back to the function would be a breaking change. Basically it is a case of the child function passing back what they received in the first place, to make it easier to chain.

README.md
ferdinando-ferreira: Heh, I agree, I used the actual use case I'm proposing this function tho. It could be changed to the one used in the test suite (a simpler replace).

Here is the real world use case this would solve (and the whole motivation for this PR): as of today here is no way to detect if the request was a POST and, now that POSTs are accepted at all (due to the `methods` recently accepted option), there is no way to actually pass the post data to the scripts.

The example contained in the readme does just that, but partially. Parse the POST body and add to the global_request var and it becomes possible to inform the scripts what the actual posted data was.

That's an overkill for the README, I agree, should I change it to the "My" -> "Modified" example I included in the tests?

michael-ciniawsky:
> 1. Inject meta tags to the index.html generated by html-webpack-plugin (like title or meta description)

For different (additional) `meta` tags based on the particular request, right ? Otherwise this should by solved by the `html-webpack-plugin` already. Please elaborate here

> 2. add watermarks to images generated on the fly

I don't know of a webpack plugin which does that, but I think it is possible to solve via plugin (but could be wrong). Plea povide an example (if possible) here

> 3. instruct this middleware not to generate certain filed so other middlewares down the line can do it instead

Ok. But this needs design e.g  a `options.content` option => `options.content.filter(fn)` (3) + eventually `options.content.transform(fn)` ([1, 2]) (currently proposed as `generateContent`) as it sounds like 3 is a different use case then 'generating' content :), where `transform` `{Function}` is more like just e.g `(url, content) => something(content)`

> Essentially this change fulfills the idea of the Server-Side Rendering option mentioned in the docs (which I assume are going away based on the Project notes) in a cleaner, more generic and more powerful manner and with no drawback (that I can imagine)

That sounds interesting, an idea for a more generic approach is appreciated here. In terms of removal nothing is set in stone yet, but the term server side rendering is confusing imho and I would like to either refactor and rename it (no concrete plans) or simply remove it. But this needs consensus, I'm voicing my personally here

ferdinando-ferreira:
> That sounds interesting, an idea for a more generic approach is appreciated here. 

I'll write a detailed explanation later but, for now, I prepared an example of what can be achieved with this potential new option.

Here are the install instruction for this sample, it is a fully featured turnkey hot reloaded Server Side rendered single file component powered environment to develop vuejs applications. 

```
git clone --single-branch -b samples https://github.com/ferdinando-ferreira/webpack-dev-middleware.git
cd webpack-dev-middleware
npm install
cd samples
cd vuessr
npm install
npm run dev
```

This little environment allows for development of Vue applications without actually having to restart the server, recompile manually or even generate the files on disk, it all happens in memory. It may not be a fully drop in replacement for nuxt but it is close enough to be useful.

I plan to write the second (unrelated, for the purpose of this exercise) half of this and write the production side of the environment to fulfill the potential of the tool but, for now, this illustrates the potential of this new option with regards to the possibilities to extend webpack-serve through webpack-dev-middleware.

Full write up later, please install and take a look.

michael-ciniawsky:
I will take a look at the example as soon as I have the time, but 

> vuejs applications

A particular framework's limitations or concerns of course influence decisions made for `webpack-dev-middleware` but it's **definitely** not sufficient to justify adding a new option. I understand this might be just used as an example here, but we need to triage the underlying concepts and find a 'as possible' generic design for the option/API **independent** from any framework design by default. I'm not happy with the name and design of this option as is. It smells like 'quick and dirty' ad-hoc solution atm (no offense, or being agressive here, it's fine to do that for **personal** usage). Every new option needs to be maintained and will likely stay for a while and potential usage of this particuar addition is also still vague and broader in scope. We need to make sure to identify the potential use case, potential issues/misuses (in conjunction with other options, `webpack-dev-server` etc etc) and how this may replace `options.serverSideRender`/improve SSR in particular in future and iterate

ferdinando-ferreira:
> no offense
None taken

> We need to make sure to identify the potential use case, potential issues/misuses (in conjunction with other options, `webpack-dev-server` etc etc) and how this may replace `options.serverSideRender`/improve SSR in particular in future and iterate
Duly noted

>  But we need to triage the underlying concepts and find a 'as possible' generic design for the option/API . I'm not happy with the name and design of this option as is. It smells like 'quick and dirty' ad-hoc solution atm 
Your assessment is correct (it is mostly brainstorming in public up until now), one of the reasons I created a separated branch instead of polluting the Pull Request even more. 

> I understand this might be just used as an example here (...) independent from any framework design by default
The vue is just an example, there is nothing specific in that particular implementation except the fact a real world example was used to illustrate the usefulness.

Test the sample provided if you can and I'll write a full write-up, think on a non provisional naming the option and try to make the case for it's inclusion.

michael-ciniawsky:
> The vue is just an example, there is nothing specific in that particular implementation except the fact a real world example was used to illustrate the usefulness.

Yep, understood and ssumed since `generateContent` is generic here already

> Test the sample provided if you can and I'll write a full write-up, think on a non provisional naming the option and try to make the case for it's inclusion.

The particular name is the least important for now, sothing lik either `options.content` or if more SSR related `options.ssr` as an `{Object}` with properties for each needed area e.g

```js
app.use(middleware(compiler, {
   content: {
      filter (url) {},
      transform (content) {},
      generate (req, res, fs) {},
      ...,
  }
}))
```

ferdinando-ferreira:
Closing this PR so I can submit the another one, synced with the current master and properly implemented

PR #334, opened 2018-08-25
fix: replace `url-join` with `path.posix.join`

### LINKED ISSUE #333: Ancestor directory names containing a leading hash not supported
* Operating System: macOS High Sierra 10.13.6
* Node Version: v10.9.0
* NPM Version: 6.4.0
* webpack version: 4.17.1
* webpack-dev-middleware Version: 3.1.3

- [x] This is a **bug**

There are several bug reports in vue-cli repository related to this bug:

- https://github.com/vuejs/vue-cli/issues/2342
- https://github.com/vuejs/vue-cli/issues/2345
- https://github.com/vuejs/vue-cli/issues/2357

To reproduce it, run the following commands:

```
mkdir "#leadinghash"
cd "#leadinghash"
npx @vue/cli create app --default
cd app
npm run serve -- --open
```

### Expected Behavior

A hello world page.

### Actual Behavior

A blank page with following error message logged in terminal:

```
URIError: Failed to decode param '/%3C%=%20BASE_URL%20%%3Efavicon.ico'
```

### Root Cause

It is actually a bug of `url-join`(used in [middleware.js](https://github.com/webpack/webpack-dev-middleware/blob/master/lib/middleware.js#L54) & [util.js](https://github.com/webpack/webpack-dev-middleware/blob/master/lib/util.js#L107)):
https://github.com/jfromaniello/url-join/issues/10

```
require('url-join').('/some/path/with/a/leading/#hash', 'index.html')
// returns '/some/path/with/a/leading#hash/index.html', the `/` before `#` is missing
```

But I think it is better to solve this issue here because:
1. Maintainers of `url-join` does not seem to have a strong willingness to fix this long-standing issue;
2. Since `url-join` is only used to join filesystem paths in `webpack-dev-middleware`, `path.posix,join` could be a sufficient replacement.

sodatea:
### Type

- [x] Fix

### Issues

- Fixes #333 

### SemVer

- [x] Patch

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -4,7 +4,6 @@ const path = require('path');
 const { parse } = require('url');
 const querystring = require('querystring');
 const parseRange = require('range-parser');
-const urlJoin = require('url-join');
 
 const HASH_REGEXP = /[0-9a-f]{10,}/;
 
@@ -92,7 +91,7 @@ module.exports = {
     if (process.platform === 'win32') {
       // Path Handling for Microsoft Windows
       if (filename) {
-        uri = urlJoin((outputPath || ''), querystring.unescape(filename));
+        uri = path.posix.join((outputPath || ''), querystring.unescape(filename));
 
         if (!path.win32.isAbsolute(uri)) {
           uri = `/${uri}`;
@@ -104,7 +103,7 @@ module.exports = {
 
     // Path Handling for all other operating systems
     if (filename) {
-      uri = urlJoin((outputPath || ''), filename);
+      uri = path.posix.join((outputPath || ''), filename);
 
       if (!path.posix.isAbsolute(uri)) {
         uri = `/${uri}`;
--- a/lib/util.js
+++ b/lib/util.js
@@ -91,7 +91,7 @@ module.exports = {
     if (process.platform === 'win32') {
       // Path Handling for Microsoft Windows
       if (filename) {
-        uri = path.posix.join((outputPath || ''), querystring.unescape(filename));
+        uri = path.posix.join(path.normalize(outputPath || ''), querystring.unescape(filename));
 
         if (!path.win32.isAbsolute(uri)) {
           uri = `/${uri}`;
--- a/lib/util.js
+++ b/lib/util.js
@@ -91,7 +91,7 @@ module.exports = {
     if (process.platform === 'win32') {
       // Path Handling for Microsoft Windows
       if (filename) {
-        uri = path.posix.join(path.normalize(outputPath || ''), querystring.unescape(filename));
+        uri = path.posix.join((outputPath || ''), querystring.unescape(filename));
 
         if (!path.win32.isAbsolute(uri)) {
           uri = `/${uri}`;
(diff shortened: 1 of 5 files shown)


michael-ciniawsky:
@sodatea Happy to see it go an use a core module instead, but windows test are currently failing with this patch https://ci.appveyor.com/project/sokra/webpack-dev-middleware/build/1.0.110/job/fxenmru3b8ujigbt

test/tests/util.js
```diff
@@ -278,41 +284,41 @@ describe('GetFilenameFromUrl', () => {
         url: '/test/windows.txt',
         outputPath: 'c:\\foo',
         publicPath: '/test',
-        // this is weird, but it's legal-ish, and what URI parsing produces
-        expected: 'c://\\foo/windows.txt'
```
michael-ciniawsky: We need ro make sure this is valid to be removed, windows paths recognize both separators to my knowledge and this teherefore will far for some folks having `/` in their paths for whatever reason: Can you elaborate on why this can be removed please ?

sodatea: https://github.com/webpack/webpack-dev-middleware/pull/276
> Where we might expect `c:\foo\bar` we instead get a URL-ized `c://\foo/bar`. Either works with the default path translation on Windows, though technically it's wrong.

If I understand correctly, the `//` is not indended to be preserved but now that it still works we didn’t bother removing it.

michael-ciniawsky: Could you try with [path.normalize](https://nodejs.org/dist/latest-v10.x/docs/api/path.html#path_path_normalize_path) ?  

```js
const filename = path.posix.join(path.normalize(...))
```

sodatea: ping @michael-ciniawsky

michael-ciniawsky: Paths are not URLs there must be a reason why `url-join` was used here instead. I didn't write this code so I'm not sure if 'this only handles file paths'  renders 💯 true.  I think the issue with `path.nomalize` was, that we would need to use the `path.posix` variant instead

```diff
- const filename = path.posix.join(path.normalize(...), ...)
+ const filename = path.posix.join(path.posix.normalize(...), ...)
```

But leave it for now until we know how to proceed here in general :)

lib/util.js
```diff
@@ -1,10 +1,10 @@
 'use strict';
 
+const path = require('path');
 const { parse } = require('url');
 const querystring = require('querystring');
 const pathabs = require('path-is-absolute');
```
michael-ciniawsky: This should be gone by now https://github.com/webpack/webpack-dev-middleware/pull/335. Could you rebase with the PR with latest `master` to ensure we won't get a bad merge here please ?

sodatea: Done 😄

michael-ciniawsky:
@sodatea Since this removes a dependency could you update the lockfile? Otherwise this should this is ready to land

lib/util.js
michael-ciniawsky: Doesn't seemed to be the case the line is still there (?)

sodatea: Ahh, sorry, forgot to `push -f` 😅
Should be fixed now.

PR #336, opened 2018-08-29
feat(middleware): add `content` option (`options.content.filter`, `options.content.transform`)

ferdinando-ferreira:
Feature

Yes

**Summary**
It's a refactoring of the functionality proposed at #330 with the requested changes implemented

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -77,36 +77,51 @@ module.exports = function wrapper(context) {
             return goNext();
           }
         }
-        if (notFound) {
-          return resolve(goNext());
-        }
-        // server content
-        let content = context.fs.readFileSync(filename);
-        content = handleRangeHeaders(content, req, res);
-
-        let contentType = mime.getType(filename);
-
-        // do not add charset to WebAssembly files, otherwise compileStreaming will fail in the client
-        if (!/\.wasm$/.test(filename)) {
-          contentType += '; charset=UTF-8';
-        }
-
-        res.setHeader('Content-Type', contentType);
-        res.setHeader('Content-Length', content.length);
-
-        const { headers } = context.options;
-        if (headers) {
-          for (const name in headers) {
-            if ({}.hasOwnProperty.call(headers, name)) {
-              res.setHeader(name, context.options.headers[name]);
-            }
+        const defaultTransform = (localRequest, fs) => new Promise((localResolve) => {
+          if (notFound) {
+            return localResolve(null);
           }
-        }
-        // Express automatically sets the statusCode to 200, but not all servers do (Koa).
-        res.statusCode = res.statusCode || 200;
-        if (res.send) res.send(content);
-        else res.end(content);
-        resolve();
+          fs.readFile(localRequest.filename, (err, content) => {
+            if (err) {
+              localResolve(null);
+            } else {
+              localResolve(content);
+            }
+          });
+        });
+        const optionTransform = (context.options && context.options.content) ? context.options.content.transform : null;
+        return defaultTransform(currentRequest, context.fs)
+          .then(content => ((!optionTransform) ? content : optionTransform(currentRequest, content, context.fs)))
+          .then((content) => {
+            if (content == null) {
+              return Promise.reject();
+            }
+            content = handleRangeHeaders(content, req, res);
+            let contentType = mime.getType(currentRequest.filename);
+
+            // do not add charset to WebAssembly files, otherwise compileStreaming will fail in the client
+            if (!/\.wasm$/.test(currentRequest.filename)) {
+              contentType += '; charset=UTF-8';
+            }
+
+            res.setHeader('Content-Type', contentType);
+            res.setHeader('Content-Length', content.length);
+
+            const { headers } = context.options;
+            if (headers) {
+              for (const name in headers) {
+                if ({}.hasOwnProperty.call(headers, name)) {
+                  res.setHeader(name, context.options.headers[name]);
+                }
+              }
+            }
+            // Express automatically sets the statusCode to 200, but not all servers do (Koa).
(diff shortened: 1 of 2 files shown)


lib/middleware.js
```diff
@@ -65,6 +65,18 @@ module.exports = function wrapper(context) {
...
+          filename,
+          notFound
+        });
+        if (context.options && context.options.content && context.options.content.filter) {
+          if (!context.options.content.filter(currentRequest)) {
+            return goNext();
+          }
+        }
         if (notFound) {
           return resolve(goNext());
```
ferdinando-ferreira: Question: why is it that every other "goNext" is a regular return but this one is a return resolve(goNext())?

lib/middleware.js
```diff
@@ -77,36 +77,51 @@ module.exports = function wrapper(context) {
...
-        // server content
-        let content = context.fs.readFileSync(filename);
-        content = handleRangeHeaders(content, req, res);
-
-        let contentType = mime.getType(filename);
+        const defaultTransform = (localRequest, fs) => new Promise((localResolve) => {
+          if (notFound) {
+            return localResolve(null);
+          }
+          fs.readFile(localRequest.filename, (err, content) => {
```
ferdinando-ferreira: Is it a problem that the readFileSync was replaced by an async one?

ferdinando-ferreira:
Missing: changing the README.md, working on it but it would gain some time to request comments on the code itself meanwhile.

test/tests/server.js
```diff
@@ -152,6 +152,114 @@ describe('Server', () => {
...
   });
 
+  describe('filters', () => {
+    before((done) => {
+      app = express();
+      const compiler = webpack(webpackConfig);
+      instance = middleware(compiler, {
+        stats: 'errors-only',
+        content: {
+          filter(currentRequest) {
```
michael-ciniawsky: `currentRequest` => `res || request`

lib/middleware.js
```diff
@@ -38,6 +38,7 @@ module.exports = function wrapper(context) {
     return new Promise(((resolve) => {
       handleRequest(context, filename, processRequest, req);
       function processRequest() {
+        let notFound;
```
michael-ciniawsky: Please elaborate further on the need for this

lib/middleware.js
```diff
@@ -60,37 +61,67 @@ module.exports = function wrapper(context) {
...
+            res.setHeader('Content-Length', content.length);
 
-        const { headers } = context.options;
-        if (headers) {
-          for (const name in headers) {
-            if ({}.hasOwnProperty.call(headers, name)) {
-              res.setHeader(name, context.options.headers[name]);
+            const { headers } = context.options;
+            if (headers) {
+              for (const name in headers) {
```
michael-ciniawsky: `name` => `header`

lib/middleware.js
```diff
@@ -60,37 +61,67 @@ module.exports = function wrapper(context) {
...
+              localResolve(content);
+            }
+          });
+        });
+        const optionTransform = (context.options && context.options.content) ? context.options.content.transform : null;
+        return defaultTransform(currentRequest, context.fs)
+          .then(content => ((!optionTransform) ? content : optionTransform(currentRequest, content, context.fs)))
+          .then((content) => {
+            if (content == null) {
+              return Promise.reject();
```
michael-ciniawsky: Should we better reject with an `{Error}` here ?

michael-ciniawsky:
> Missing: changing the README.md, working on it but it would gain some time to request comments on the code itself meanwhile.

Yep, let's discuss the implementation first and once approve you can add the docs then :)

lib/middleware.js
michael-ciniawsky: Good question but I don't know the answer, try to remove the `resolve` and see what's happening (if something happens). Please add a comment that this is something to investigate and clearify :)

evilebottnawi:
@michael-ciniawsky can we describe this feature in main post briefly, it is hard to read a lot of posts?

ferdinando-ferreira:
> can we describe this feature in main post briefly, it is hard to read a lot of posts?

Here is the shortest summary:
 * `options.content.filter`: receives a function with the following arguments: `(request)`. This function is called once for each resource being served by webpack-dev-middleware. Returning `false` makes `webpack-dev-middleware` not to serve said file and pass it along to the next middleware.
 * `options.content.transform`: receives a function with the following arguments: `(request, content, memoryFs)`. This function is called once for each resource being served by webpack-dev-middleware and must return a promise whose `resolve` contains the Buffer with the actual content to be served and whose `reject` makes `webpack-dev-middleware` not to serve said file and pass it along to the next middleware.

The added tests contain examples of use cases for these options and, in general, they supercede what's being offered by `options.serverSideRender`, but in a more generic way.

lib/middleware.js
sokra: `fs` is a in memory filesystem so it doesn't matter

evilebottnawi:
@ferdinando-ferreira 
> sokra [6:10 PM]
> hmm.. not sure if this is needed. It's a middleware, so you could create another middleware to transform the responses. And anyway transformations to assets can be done as webpack plugin...

What do you think about this solution?

ferdinando-ferreira:
> I can propose a smaller change to middleware.js that passes fs along with webpackStats to the next middleware in line 

Done, PR #337. Both PRs solve the same issue, this one in a more generic fashion, the other one contingent on ServerSideRender flag not going away.

ferdinando-ferreira:
Rescinding this PR, #337 is enough as long as ServerSideRequest doesn't get dropped.

michael-ciniawsky:
We will definitely keep it until we found a equally sufficient replacement for SSR workflows. I will review #337 asap, thx for your efforts  on this so far :)

PR #340, opened 2018-09-23
Remove loud-rejection

### LINKED ISSUE #339: Swallows all async errors in server context
### Environment

It believe this issue applies to all versions. Nevertheless here is my env:

* Operating System: MacOS 10.13.6
* Node Version: v10.9.0
* NPM Version: 6.2.0
* webpack version: ^4.17.1
* webpack-dev-middleware Version: ^3.2.0

- [x] This is a **bug**

### Code

https://github.com/HaNdTriX/webpack-dev-middleware-bug

### Expected Behavior

webpack-dev-middleware should not modify the global `uncaughtException` behaviour.

### Actual Behavior

webpack-dev-middleware swallows all `uncaughtException`s and **only** logs them when the process exits. This never happens since this middleware runs in a server context.
This is critical since it also swallows `uncaughtException` that are not related to webpack, express & the compiled code.

1. Run the following example: https://github.com/HaNdTriX/webpack-dev-middleware-bug

```bash
$ git clone git@github.com:HaNdTriX/webpack-dev-middleware-bug.git
$ cd webpack-dev-middleware-bug
$ npm install
$ npm start
```

2. Open:

https://github.com/HaNdTriX/webpack-dev-middleware-bug/blob/master/index.js#L10-L25

### Explanation

webpack-dev-middleware uses a module called https://github.com/sindresorhus/loud-rejection.
This module changes the global `uncaughtException` behaviour.

@sindresorhus quotes in the readme:

> Use this in top-level things like tests, CLI tools, apps, etc, but not in reusable modules.

### Solution

Remove https://github.com/sindresorhus/loud-rejection from this module. I don't think we need it anyway.
I currently can't think of usecases where this is useful. Please correct me if I am wrong.

HaNdTriX:
[loud-rejection](https://github.com/sindresorhus/loud-rejection) modifies the global promise/async rejection behaviour so no uncaught rejections will be swallowed.

bugfix

TBD

**Summary**

See: #339 

This PR introduces a breaking change since uncaught promises might end the process now.

**Other information**

Closes #339

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,8 +7,6 @@ const reporter = require('./lib/reporter');
 const { setFs, toDisk } = require('./lib/fs');
 const { getFilenameFromUrl, noop, ready } = require('./lib/util');
 
-require('loud-rejection/register');
-
 const defaults = {
   logLevel: 'info',
   logTime: false,
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.51",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.17.5",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.1",
+        "lodash": "4.17.10",
+        "source-map": "0.5.7",
+        "trim-right": "1.0.1"
       }
     },
     "@babel/helper-function-name": {
@@ -61,9 +61,9 @@
       "integrity": "sha1-6IRK4loVlcz9QriWI7Q3bKBtIl0=",
       "dev": true,
       "requires": {
-        "chalk": "^2.0.0",
-        "esutils": "^2.0.2",
-        "js-tokens": "^3.0.0"
+        "chalk": "2.4.1",
+        "esutils": "2.0.2",
+        "js-tokens": "3.0.2"
       }
     },
     "@babel/parser": {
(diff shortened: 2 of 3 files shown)


SpaceK33z:
@evilebottnawi I can't come up with a good reasoning as to why this would be a breaking change. I see this only as a bug, especially since in `loud-rejection`'s README it is mentioned it should not be used in packages. I think fixing this is in a minor release rather than waiting for a major release (which can take months) is better for more users.

evilebottnawi:
@SpaceK33z Agree, i think we should release patch version after this merge

SpaceK33z:
@evilebottnawi Fixed already, was writing the changelog by hand because I saw the new `npm run release` script too late (wasn't there last time I published a while ago 😄)

HaNdTriX:
Wow that was fast. Thanks a lot 🙏!

michael-ciniawsky:
Why was this released as a `minor` version bump it should either have been a `patch` (`v3.3.2`) or a major (`v4.0.0`)? The `CHANGELOG` is also compelete missing https://github.com/webpack/webpack-dev-middleware/blob/master/CHANGELOG.md as it seems (I fix it manually this time, please **don't** do anything). Could you guys please just leave the PR(s) open if you are not familiar with the release 'process' (yet) (I merge approved PR's and release everything once per day normally), or ask if in doubt? Everytime someone else cuts a release in one of `webpack/webpack-contrib` repos it is more or less done **wrong**

SpaceK33z:
@michael-ciniawsky this was a `minor` version bump because I found it too tricky to publish as patch, yet unnecessary to release as a `major`. Maybe in hind sight not the best option, but too late.

Anyway, as I said above here already I didn't know there was a `npm run release` script. I've been maintaining this repository for two years now, and suddenly there is a release script I only could've known if I super carefully read every line of the `package.json`. If you want to prevent people from doing it wrong, please block using `npm version` directly. I'm sure there are ways to do that. Without that, this mistake will be repeated and repeated by me and others.

PR #136, opened 2016-10-04
Added charset=UTF-8 to content-type to fix encoding issues.

callumacrae:
Fixes https://github.com/webpack/webpack-dev-server/issues/432

This hasn't broken anything for me.

PR #351, opened 2018-12-13
Only add charset to content-type header if defined in mime db

### LINKED ISSUE #350: Unwanted charset=UTF-8 always added to custom mime types
* Operating System: macOS 10.14
* Node Version: 10.4.1
* NPM Version: 6.4.1
* webpack version: 4.27.1

- [x] This is a **feature** request
- [x] This is a **bug**

Since https://github.com/webpack/webpack-dev-middleware/pull/136, `charset=UTF-8` is always appended to the Content-Type header.  If I define custom mime types, I believe that this shouldn't be appended.  

For example, mobile safari can render 3d/AR models in a native viewer when a file is served with ` model/vnd.pixar.usd`. When setting the option `mimeTypes: { 'model/vnd.pixar.usd', ['usdz'] }`, the server will return: `Content-Type: model/vnd.pixar.usd; charset=UTF-8`.  The addition of the charset causes the browser to incorrectly interpret the mime type.

This is already special-cased for wasm.

### Expected Behavior
charset=UTF-8 is not appended to every Content-Type header when defining custom mimeTypes.

### Actual Behavior
charset=UTF-8 is always appended

I can make a PR if you agree on the proposed behavior.

gpoitch:
bugfix/functionality change

Yes

**Summary**
Closes #350 
`charset=utf-8` was being appended to the content-type header of every mime type regardless if it required it (#136).  This PR uses mime-db to look up the charset for each mime type and add it appropriately.

Don't think so unless someone was relying on the previously incorrect behavior.

**Other information**
- I removed Content-Type tests from things that expect a 404. Not sure how/what was setting it.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -3,6 +3,7 @@
 const path = require('path');
 
 const mime = require('mime');
+const mimeDb = require('mime-db');
 
 const DevMiddlewareError = require('./DevMiddlewareError');
 const {
@@ -11,10 +12,6 @@ const {
   handleRequest,
   ready,
 } = require('./util');
-
-// Do not add a charset to the Content-Type header of these file types
-// otherwise the client will fail to render them correctly.
-const NonCharsetFileTypes = /\.(wasm|usdz)$/;
 
 module.exports = function wrapper(context) {
   return function middleware(req, res, next) {
@@ -97,14 +94,17 @@ module.exports = function wrapper(context) {
 
         content = handleRangeHeaders(content, req, res);
 
-        let contentType = mime.getType(filename) || '';
-
-        if (!NonCharsetFileTypes.test(filename)) {
-          contentType += '; charset=UTF-8';
-        }
-
         if (!res.getHeader('Content-Type')) {
-          res.setHeader('Content-Type', contentType);
+          const contentType = mime.getType(filename);
+          const mimeInfo = mimeDb[contentType];
+          const charset = mimeInfo && mimeInfo.charset;
+
+          if (contentType) {
+            res.setHeader(
+              'Content-Type',
+              contentType + (charset ? `; charset=${charset}` : '')
+            );
+          }
         }
 
         res.setHeader('Content-Length', content.length);
--- a/package-lock.json
+++ b/package-lock.json
@@ -6902,8 +6902,7 @@
     "mime-db": {
       "version": "1.38.0",
       "resolved": "https://registry.npmjs.org/mime-db/-/mime-db-1.38.0.tgz",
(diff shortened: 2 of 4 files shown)


gpoitch:
Only thing _maybe_ breaking is 'text/html' no longer has the charset.  I assume because it could be multiple different values.  It is usually set with the meta tag `<meta charset="utf8">`.

Test is failing because package-lock is out of sync.  Can make a separate PR to sync it on master if you'd like.

gpoitch:
I could special case text/html to add it so this isn't potentially breaking.

evilebottnawi:
@gpoitch feel free to send

lib/middleware.js
```diff
@@ -70,10 +71,13 @@ module.exports = function wrapper(context) {
         content = handleRangeHeaders(content, req, res);
 
         let contentType = mime.getType(filename);
+        const mimeInfo = mimeDb[contentType];
+        const charset = (mimeInfo && mimeInfo.charset)
+          // Special-cased to prevent possible breaking change
```
evilebottnawi: Add `TODO` about what we need remove this in next major :+1:

lib/middleware.js
```diff
@@ -70,10 +71,14 @@ module.exports = function wrapper(context) {
...
         let contentType = mime.getType(filename);
-
-        // do not add charset to WebAssembly files, otherwise compileStreaming will fail in the client
-        if (!/\.wasm$/.test(filename)) {
-          contentType += '; charset=UTF-8';
+        const mimeInfo = mimeDb[contentType];
+        const charset = (mimeInfo && mimeInfo.charset)
+          // Special-cased text/html to prevent possible breaking change.
+          // TODO: remove the following line in next major release.
+          || (contentType === 'text/html' && 'UTF-8');
```
clshortfuse: `&& 'UTF-8'` always evaluates to `true`;

gpoitch: It evaluates to the literal string 'UTF-8' to assign it to the charset variable.

gpoitch:
@evilebottnawi good to go

evilebottnawi:
@gpoitch A am little afraid about breaking change, maybe we can merge this and release after holidays? For fast fixing/reverting if problem will be for many developers?

gpoitch:
@evilebottnawi There is a special condition to avoid the breaking change.

As an alternative, we could maintain our own blacklist (.wasm, .usdz so far) instead of using mime-db.  That would leave everything else as-is.

gpoitch:
Hmm been thinking about it.  I think maybe just blacklisting `usdz` like `wasm` is on master would be much less disruptive. If anymore types come up we could reconsider this.

gpoitch:
@evilebottnawi Here's a simpler/safer alternative PR: https://github.com/webpack/webpack-dev-middleware/pull/357

gpoitch:
Closing for #357

evilebottnawi:
Close in favor https://github.com/webpack/webpack-dev-middleware/pull/357

PR #353, opened 2018-12-18
fix: use mkdirp package instead of implicit webpack dependency

connor4312:

bugfix

no (but no observable changes, and tests pass)

**Summary**

I'm working on perhaps a somewhat exotic usage of this plugin. I noticed, when getting "module not found" errors, that it implicitly depends on `webpack` core, but only for one function (mkdirp) which simply [delegated](https://github.com/webpack/webpack/blob/master/lib/node/NodeOutputFileSystem.js#L13_ to the `mkdirp`. Pull that in directly here.

No

**Other information**

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -2,12 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
+const mkdirp = require('mkdirp');
 const MemoryFileSystem = require('memory-fs');
 const { colors } = require('webpack-log');
-const NodeOutputFileSystem = require('webpack/lib/node/NodeOutputFileSystem');
 const DevMiddlewareError = require('./DevMiddlewareError');
-
-const { mkdirp } = new NodeOutputFileSystem();
 
 module.exports = {
   toDisk(context) {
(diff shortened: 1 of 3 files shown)


connor4312:
The test case is `require`'ing the webpack-dev-middleware without webpack installed in the module doing the requiring. Specifically for me, I have a small alternatively CLI that is made to be separately-installing and contains a development server using this middleware. Webpack is usually installed in the consumer's folder, and we trigger the build by manually requiring it, but the CLI itself doesn't have webpack installed.

I'm sort of confused by this statement:

> we can't use external file system package for file system

Previously the code was directly importing NodeOutputFileSystem and pulling off the mkdir, which was reexported right from the mkdirp module.

[](https://github.com/webpack/webpack/blob/master/lib/node/NodeOutputFileSystem.js#L9-L13)

Unless there's some kind of monkey-patching going on, I'm unsure how this PR changes any code or functionality, except for not breaking if webpack is not installed

evilebottnawi:
Package inside webpack fs can be changed in future, so it is better always use `webpack` filesystem implementation

connor4312:
I agree if you were using the filesystem implementation wholesale, but when the sole function this middleware is using is mkdirp, it's probably pretty safe to assume that that's not going to fundamentally change; GNU mkdirp is decades old.

PR #357, opened 2019-01-07
Don't add charset to certain file types

### LINKED ISSUE #350: Unwanted charset=UTF-8 always added to custom mime types

gpoitch:
Bugfix. #350
Less aggressive alternative to PR #351 

Yes

**Summary**
Creates a variable to special case more file types that should not automatically have a charset added to their content-type header.  This was already the case for wasm and this also adds usdz.

No

**Other information**
I originally attempted to look up each filetype in mime-db (#351) to determine if it should receive a charset.  However, that only looks up if the file type has a "default" charset defined by the spec which is not necessarily good for friendly out-of-the box support and could have introduced breaking changes.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -5,6 +5,10 @@ const mime = require('mime');
 const DevMiddlewareError = require('./DevMiddlewareError');
 const { getFilenameFromUrl, handleRangeHeaders, handleRequest, ready } = require('./util');
 
+// Do not add a charset to the Content-Type header of these file types
+// otherwise the client will fail to render them correctly.
+const NonCharsetFileTypes = /\.(wasm|usdz)$/;
+
 module.exports = function wrapper(context) {
   return function middleware(req, res, next) {
     // fixes #282. credit @cexoso. in certain edge situations res.locals is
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Previous PR was right, just need major release for this, but this we can merge as bugfix, please reopen old PR for future major release, thanks

evilebottnawi:
Released https://github.com/webpack/webpack-dev-middleware/releases/tag/v3.5.2

Issue #359, opened 2019-01-15
ConcurrentCompilationError with lazy: true when multiple entry points present


MonochromeChameleon:
* Operating System: macOS High Sierra 
* Node Version: 10.15.0
* NPM Version: 6.5.0
* webpack version: 4.28.4
* webpack-dev-middleware Version: 3.5

- [ ] This is a **feature** request
- [X] This is a **bug**

### Code

```js
// webpack.config
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

module.exports = {
  mode: 'development',
  plugins: [new MiniCssExtractPlugin()],
  module: {
    rules: [{
      test: /\.css$/,
      use: [MiniCssExtractPlugin.loader, 'css-loader'],
    }],
  },
};
```

```js
// ./src/main.js
import './style.css';

document.write('it works');
```

```css
/* ./src/style.css */
* {
    font-size: xx-large;
}
```

```html
<!-- ./src/index.html -->
<html lang="en">
<head>
  <link rel="stylesheet" href="./main.css"/>
  <script src="./main.js"></script>
</head>
</html>
```

```js
// ./server.js
const express = require('express');
const webpack = require('webpack');
const WebpackDevMiddleware = require('webpack-dev-middleware');
const webpackConfig = require('./webpack.config');

const lazy = process.argv.reverse()[0] === '--lazy';
const app = express();

app.use(WebpackDevMiddleware(webpack(webpackConfig), { lazy }));
app.use(express.static('./src'));

app.listen(3000, () => console.log('app started'));
```

### Expected Behavior

All assets should be compiled and served, allowing the page to load

### Actual Behavior

Behaviour is as expected when webpack-dev-middleware is configured with lazy: false.
When lazy: true is supplied as an option, the two requests for css & js both trigger recompilation, resulting in a concurrent compilation error.

### How can we reproduce the behavior?

Create any project with two compiled assets requested in parallel, served by webpack-dev-middleware with { lazy: true }

### Possible solution

Change lib/util.js around line 152 from:
```js
// in lazy mode, rebuild on bundle request
if (context.options.lazy && (!context.options.filename || context.options.filename.test(filename))) {
  context.rebuild();
}
```
to
```js
// in lazy mode, rebuild on bundle request
if (context.options.lazy && !context.compiler.running && (!context.options.filename || context.options.filename.test(filename))) {
  context.rebuild();
}
```

evilebottnawi:
Why you use `lazy`?

[...]

PR #360, opened 2019-01-15
fix(middleware): do not trigger multiple concurrent rebuilds in lazy mode

MonochromeChameleon:

bugfix

Yes

**Summary**

Addresses https://github.com/webpack/webpack-dev-middleware/issues/359

I don't think so

**Other information**

It may be that the watchoptions config means that this isn't necessary, in which case feel free to reject :)

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -149,7 +149,7 @@ module.exports = {
 
   handleRequest(context, filename, processRequest, req) {
     // in lazy mode, rebuild on bundle request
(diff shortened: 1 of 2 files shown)


evilebottnawi:
I agree with @evenstensberg

MonochromeChameleon:
My colleague with windows said:

> I've tried those watch options before... they're buggy on docker/windows, where they recognize change (yay) via polling (egh), but they reload the *old* version of the file (boo)
> lazy was the only thing that worked
> very annoying

evilebottnawi:
Closing due inactivity

PR #361, opened 2019-01-16
Remove querystring from filenames when writing to disk

GregWeil:

This is a bug fix to bring file writing behavior closer to that of webpack and the WriteFilePlugin.

Yes

**Summary**

Webpack and the WriteFilePlugin cut off filenames at the first ?, so you can use filenames like `'[name].js?[contenthash]'` without writing hashes to your filesystem. This updates the dev middleware to do the same when `writeToDisk` is enabled.

If you were using writeToDisk with a ? in `output.filename` then you won't end up with a ? in filenames on disk any more. If you pass a filter function to `writeToDisk` then that function will also stop getting the querystring.

**Other information**

Here's where querystrings get cut off in webpack and the writefileplugin:
https://github.com/webpack/webpack/blob/v4.28.4/lib/Compiler.js#L321
https://github.com/gajus/write-file-webpack-plugin/blob/v4.4.0/src/WriteFileWebpackPlugin.js#L174

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -26,8 +26,9 @@ module.exports = {
         for (const assetPath of Object.keys(assets)) {
           const asset = assets[assetPath];
           const source = asset.source();
-          const isAbsolute = path.isAbsolute(assetPath);
-          const writePath = isAbsolute ? assetPath : path.join(outputPath, assetPath);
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Released in https://github.com/webpack/webpack-dev-middleware/releases/tag/v3.5.1

PR #366, opened 2019-02-09
use memfs instead of memory-fs

heygrady:
Uses [`memfs`](https://www.npmjs.com/package/memfs) instead of `memory-fs` for the in-memory `outputFileSystem`. `memfs` appears to be actively maintained, while `memory-fs` hasn't had any meaningful updates in a few years.

This allows people to use out-of-the-box webpack-dev-server for novel use-cases where having a fully API-compliant filesystem would be beneficial. Related: https://github.com/webpack/memory-fs/issues/43

I ran into issues adding my own server-side-rendering middleware into webpack-dev-server that needed to `require` files from the `outputFileSystem`. Doing this with memory-fs was difficult, time consuming and seemed not to work very reliably. My alternative was to write to disk, which means 30 second re-builds.

Interestingly, `memfs` is part of a suite of tools that enables useful stuff like [patching `require` to import modules from an in-memory file system](https://github.com/streamich/fs-monkey/blob/master/docs/api/patchRequire.md).

This was a simple drop-in replacement.

- [x] swap memory-fs with memfs
- [ ] move the `join` and `normalize` methods from memory-fs to "a better place"
- [x] All tests pass

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -2,10 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
-const { fs: memfs } = require('memfs');
-// borrow join and normalize from memory-fs
-memfs.join = require('memory-fs/lib/join');
-memfs.normalize = require('memory-fs/lib/normalize');
+const { Volume } = require('memfs');
+// TODO: move these two functions to a better place
+const memoryFsJoin = require('memory-fs/lib/join');
+const memoryFsNormalize = require('memory-fs/lib/normalize');
 const { colors } = require('webpack-log');
 const NodeOutputFileSystem = require('webpack/lib/node/NodeOutputFileSystem');
 const DevMiddlewareError = require('./DevMiddlewareError');
@@ -63,12 +63,14 @@ module.exports = {
 
     let fileSystem;
     // store our files in memory
-    const isMemoryFs = !compiler.compilers && compiler.outputFileSystem === memfs;
+    const isMemoryFs = !compiler.compilers && compiler.outputFileSystem instanceof Volume;
 
     if (isMemoryFs) {
       fileSystem = compiler.outputFileSystem;
     } else {
-      fileSystem = memfs;
+      fileSystem = new Volume();
+      fileSystem.join = memoryFsJoin;
+      fileSystem.normalize = memoryFsNormalize;
       compiler.outputFileSystem = fileSystem;
     }
 
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -2,7 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
-const MemoryFileSystem = require('memory-fs');
+const { fs: memfs } = require('memfs');
+// borrow join and normalize from memory-fs
+memfs.join = require('memory-fs/lib/join');
+memfs.normalize = require('memory-fs/lib/normalize');
 const { colors } = require('webpack-log');
 const NodeOutputFileSystem = require('webpack/lib/node/NodeOutputFileSystem');
 const DevMiddlewareError = require('./DevMiddlewareError');
@@ -60,12 +63,12 @@ module.exports = {
 
     let fileSystem;
     // store our files in memory
-    const isMemoryFs = !compiler.compilers && compiler.outputFileSystem instanceof MemoryFileSystem;
+    const isMemoryFs = !compiler.compilers && compiler.outputFileSystem === memfs;
 
     if (isMemoryFs) {
       fileSystem = compiler.outputFileSystem;
     } else {
-      fileSystem = new MemoryFileSystem();
+      fileSystem = memfs;
       compiler.outputFileSystem = fileSystem;
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -2543,6 +2543,11 @@
       "integrity": "sha1-ewUhjd+WZ79/Nwv3/bLLFf3Qqkk=",
       "dev": true
     },
+    "fast-extend": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/fast-extend/-/fast-extend-0.0.2.tgz",
+      "integrity": "sha1-9exCz0C5Rg9SGmOH37Ut7u1nHb0="
+    },
     "fast-json-stable-stringify": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/fast-json-stable-stringify/-/fast-json-stable-stringify-2.0.0.tgz",
@@ -2820,6 +2825,11 @@
       "requires": {
         "minipass": "^2.2.1"
       }
+    },
+    "fs-monkey": {
+      "version": "0.3.3",
+      "resolved": "https://registry.npmjs.org/fs-monkey/-/fs-monkey-0.3.3.tgz",
+      "integrity": "sha512-FNUvuTAJ3CqCQb5ELn+qCbGR/Zllhf2HtwsdAtBi59s1WeCjKMT81fHcSu7dwIskqGVK+MmOrb7VOBlq3/SItw=="
     },
     "fs-write-stream-atomic": {
       "version": "1.0.10",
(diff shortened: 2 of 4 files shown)


evilebottnawi:
We need fix it webpack first https://github.com/webpack/webpack/blob/b934e26da8b65bdf36b3c71141d0c2d8397d9f44/package.json#L22

https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/MemoryOutputFileSystem.js

release new patch version and merge this pr

lib/fs.js
```diff
@@ -2,7 +2,10 @@
 
 const fs = require('fs');
 const path = require('path');
-const MemoryFileSystem = require('memory-fs');
+const { fs: memfs } = require('memfs');
+// borrow join and normalize from memory-fs
+memfs.join = require('memory-fs/lib/join');
+memfs.normalize = require('memory-fs/lib/normalize');
```
evilebottnawi: It is not good solution, we don't need two memory-fs package

evilebottnawi:
Also can you provide example where
> I ran into issues adding my own server-side-rendering middleware into webpack-dev-server that needed to `require` files from the `outputFileSystem`. Doing this with memory-fs was difficult, time consuming and seemed not to work very reliably. My alternative was to write to disk, which means 30 second re-builds.

lib/fs.js
heygrady: Those functions appear to be unrelated to the file system itself and could live somewhere else.

evilebottnawi: I know, but it is bad have two memory fs package, we should move these function to separate repo or move to `memfs`

heygrady:
I was hacking my custom middleware into the webpack-dev-server like this:

## `start.js`

```js
// start.js
// use custom serverSideRenderMiddleware middleware with webpack-dev-server
const serverSideRenderMiddleware = require('./ssr-middleware')

// Copy the defaultFeatures logic from webpack-dev-server. We need to hack the
// default features in order to ensure that our custom serverSideRender
// middleware always runs after contentBaseFiles but before the other default
// features. Running "after" last will result in the public index always being
// rendered (for some reason).
// TODO: is there a better way?
const createDefaultFeatures = options => {
  const { after, contentBase } = options;
  const defaultFeatures = ['before', 'setup', 'headers', 'middleware'];
  if (options.proxy) {
    defaultFeatures.push('proxy', 'middleware');
  }
  if (contentBase !== false) {
    defaultFeatures.push('contentBaseFiles');
  }
  // Ensure "after" runs after "middleware" and "contentBaseFiles" but before everything else
  if (after) {
    defaultFeatures.push('after');
  }
  if (options.watchContentBase) {
    defaultFeatures.push('watchContentBase');
  }
  if (options.historyApiFallback) {
    defaultFeatures.push('historyApiFallback', 'middleware');
    if (contentBase !== false) {
      defaultFeatures.push('contentBaseFiles');
    }
    // Ensure "after" runs after "middleware" and "contentBaseFiles" but before everything else
    if (after) {
      defaultFeatures.push('after');
    }
  }
  defaultFeatures.push('magicHtml');
  // NOTE: contentBaseIndex is the devil 😈. *Never* enable it.
  // if (contentBase !== false) { defaultFeatures.push('contentBaseIndex'); }
  // compress is placed last and uses unshift so that it will be the first middleware used
  if (options.compress) {
    defaultFeatures.unshift('compress');
  }
  return defaultFeatures;
};

let afterCounter = 0;
const serverConfig = {
  // ... whatever settings you like
  serverSideRender: true,
  after(app) {
    if (afterCounter === 0) {
      // This allows us to render the HTML on the server. We don't need this
      // unless we're doing SSR.
      //
      // serverSideRenderMiddleware should be an express middleware function
      // that uses `res.send` to deliver the rendered HTML.
      //
      // You may wish to review the webpack-dev-middleware for more
      // information on how to handle webpackStats in your middleware.
      // https://github.com/webpack/webpack-dev-middleware#server-side-rendering
      app.use(serverSideRenderMiddleware);
    }
    // Because of the way we're abusing after, it appears multiple times in
    // the sequence. Here we increment a counter to better control which after
    // we're trying to target. Above you can see that we want to inject the
    // ssr middleware on the first after. The after callback runs three times.
    afterCounter++;
  },
};

// While webpack-dev-middleware supports a serverSideRender option,
// webpack-dev-server does not. We're passing in the serverSideRender
// option from webpackDevServer.config.js but there's no good way to have
// our reactApp middleware run in the right place. So, we need to override
// the features option to include our "after" middleware in the right spots.
if (serverConfig.serverSideRender) {
  // NOTE: this will break if you're already using the features option.
  if (serverConfig.features) {
    console.warn('The features option has been replaced to enable ssr.');
  }
  serverConfig.features = createDefaultFeatures(serverConfig);
}

// create the dev server
const config = { /* ... your client-side webpack config */ }
const configServer = { /* ... your server-side webpack config */ }
const compiler = webpack([config, configServer]);
const devServer = new WebpackDevServer(compiler, serverConfig);

// start the dev server
const PORT = parseInt(process.env.PORT, 10) || 3000;
const HOST = process.env.HOST || '0.0.0.0';
devServer.listen(PORT, HOST, err => {
  if (err) {
    return console.log(err);
  }
  console.log('Starting the development server');
});

['SIGINT', 'SIGTERM'].forEach(function(sig) {
  process.on(sig, function() {
    devServer.close();
    process.exit();
  });
});
```

## `ssr-middleware.js`

```js
// ssr-middleware.js
const path = require('path')
const { wrap } = require('async-middleware')

const fs = require('fs')
const { Union } = require('unionfs')
const { patchRequire } = 'fs-monkey'

// TODO: not positive these paths are correct (pseudo-code)
const mainPath = path.join(__dirname, '..', 'dist/server/main.js')
const statsPath = path.join(__dirname, '..', 'dist/static/react-loadable.json')
const manifestPath = path.join(__dirname, '..', 'dist/static/asset-manifest.json')

const mountMemoryFileSystem = async (res) => {
  const { stats } = res.locals.webpackStats
  // TODO: is this correct? (psuedo-code)
  const { outputFileSystem: clientFs } = stats[0].compilation.compiler
  const { outputFileSystem: serverFs } = stats[1].compilation.compiler
  const ufs = new Union()
  // allow us to require from all three volumes as if they were one
  ufs.use(clientFs).use(serverFs).use(fs)
  patchRequire(ufs)
}

const initMiddleware = async (res) => {
  // the main app exports a createMiddleware function
  const { createMiddleWare, preloadAll } = requireNoCache(mainPath)
  // we also need assets from the client build
  const stats = requireNoCache(statsPath)
  const manifest = requireNoCache(manifestPath)
  await preloadAll()
  const middleware = createMiddleWare(stats, manifest)
  return middleware
}

// https://stackoverflow.com/a/16060619
const requireNoCache = (module) => {
  delete require.cache[require.resolve(module)]
  return require(module)
}

const isMemoryFileSystem = (res) => {
  const stats = res.locals.webpackStats.stats[1]
  const { outputFileSystem } = stats.compilation.compiler
  return outputFileSystem instanceof Volume
}

// We don't really want to continue to the other middleware
const fakeNext = () => undefined

module.exports = wrap(async (req, res, next) => {
  if (isMemoryFileSystem(res)) {
    // TODO: this would remount the filesystem on every request
    // should probably check if that's necessary
    mountMemoryFileSystem(res)
  }
  const middleware = await initMiddleware(res)
  middleware(req, res, fakeNext)
})

```

heygrady:
https://github.com/streamich/memfs/issues/325#issuecomment-462072279

hinell:
This is nice PR but I got far better alternative I believe. #370

evilebottnawi:
@hinell yep, https://github.com/webpack/webpack-dev-middleware/pull/370 better

If somebody have problem with memory-fs please open issue

laynef:
> @hinell yep, #370 better
> 
> If somebody have problem with memory-fs please open issue

I am currently having issues with maxing out memory, I've written my files to disk and that is not helping

hinell:
@laynef It's better to open a separate issue on that. It would be also nice to elaborate it with details. So far everything it worked smoothly for me.

evilebottnawi:
You can configure fs using options, maybe we revisit default memory fs in future, no need right now

evilebottnawi:
Anyway thank you for the PR

PR #370, opened 2019-02-16
Configurable file system via options.fs

hinell:
* fs option which can be used to configure non-default file system
(it is forced to use `memory-fs` instead currently)

**What are use cases?**
* The same as for middleware, but with opportunity to use other memory file systems (like [memfs](https://github.com/streamich/memfs))

* adds couple of tests

**Summary**
* Solves longstanding problem of not having opportunity to use other non-memory-fs libraries and configure it manually either via `webpackCompiler.outputFileSystem` or `middleware(..., ..., { fs: fileSystem }`

* Negative

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ app.use(middleware(compiler, {
 app.listen(3000, () => console.log('Example app listening on port 3000!'))
 ```
 
+
 ## Options
 
 The middleware accepts an `options` Object. The following is a property reference
@@ -210,7 +211,7 @@ please see the [webpack documentation](https://webpack.js.org/configuration/watc
 Type: `Boolean|Function`  
 Default: `false`
 
-If true, the option will instruct the module to write files to the configured
+If `true`, the option will instruct the module to write files to the configured
 location on disk as specified in your `webpack` config file. _Setting
 `writeToDisk: true` won't change the behavior of the `webpack-dev-middleware`,
 and bundle files accessed through the browser will still be served from memory._
@@ -230,6 +231,14 @@ of `true` _will_ write the file to disk. eg.
   }
 }
 ```
+
+### fs
+Type: `Object`  
+Default: `MemoryFileSystem`
+
+Set the default file system which will be used by webpack as primary destination of generated files. Default is set to webpack's default file system: [memory-fs](https://github.com/webpack/memory-fs). This option isn't affected by the [writeToDisk](#writeToDisk) option.
+
+**Note:** As of 3.5.x version of the middleware you have to provide `.join()` method to the `fs` instance. This can be done simply by using `path.join`.
 
 ## API
 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ app.use(middleware(compiler, {
 app.listen(3000, () => console.log('Example app listening on port 3000!'))
 ```
 
-
 ## Options
 
 The middleware accepts an `options` Object. The following is a property reference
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Please a

evilebottnawi:
Please accept CLA, please describe use case

hinell:
@evilebottnawi Done.

README.md
```diff
@@ -231,6 +232,14 @@ of `true` _will_ write the file to disk. eg.
...
 ```
 
+### fs
+Type: `Object`  
+Default: `MemoryFileSystem`
+
+Set the default file system which will be used by webpack as primary destination of generated files. Default is set to webpack's default file system: [memory-fs](https://github.com/webpack/memory-fs). This option isn't affected by the [writeToDisk](#writeToDisk) option.
+
+**Note:** As of 3.5.x version of the middleware you have to provide `.join()` method to the `fs` instance. This can be done simply by using `path.join`.
+
```
evilebottnawi: Can we provide example how configure custom fs here?

README.md
```diff
@@ -55,6 +55,7 @@ app.use(middleware(compiler, {
 app.listen(3000, () => console.log('Example app listening on port 3000!'))
 ```
 
+
```
evilebottnawi: Avoid extra new line

test/tests/file-system.js
```diff
@@ -37,6 +37,40 @@ describe('FileSystem', () => {
...
+      done();
+    });
+
+    it('should assign fs to the compiler.outputFileSystem', (done) => {
+      const instance = middleware(compiler, { fs });
+
+      assert.equal(compiler.outputFileSystem, fs);
+      instance.close(done);
+    });
+    it('should go safely when compiler.outputFileSystem is assigned by fs externally', (done) => {
```
evilebottnawi: Add newline before `it`

hinell: Done.

evilebottnawi:
Wait CI green and merge :+1:

lib/fs.js
```diff
@@ -60,9 +60,21 @@ module.exports = {
...
-    const isMemoryFs = !compiler.compilers && compiler.outputFileSystem instanceof MemoryFileSystem;
+    const isConfiguredFs = context.options.fs;
+    const isMemoryFs = !isConfiguredFs
+      && !compiler.compilers
+      && compiler.outputFileSystem instanceof MemoryFileSystem;
 
-    if (isMemoryFs) {
+    if (isConfiguredFs) {
+      const { fs } = context.options;
+      if (typeof fs.join !== 'function') {
```
evilebottnawi: Just questions, can you find place where we use `fs.join`? Looks like we should remove this/deprecated from code base.

hinell: It is not used by the `webpack-middleware` but rather by `wepback` instead. Here it is as a precaution to avoid digging up webpack's erros. If you attempt to provide `fs` without `join()` webpack will complain about it when accessing `webpack.outputFileSystem`.

I can switch to Russian to explain more if you like.

evilebottnawi: > I can switch to Russian to explain more if you like.

No need

> It is not used by the `webpack-middleware` but rather by `wepback` instead. Here it is as a precaution to avoid digging up webpack's erros. If you attempt to provide `fs` without `join()` webpack will complain about it when accessing `webpack.outputFileSystem`.

I think we should open issue in webpack and remove all `fs.join` in `webpack@5`, `join` is not part of `fs` and looks very dirty and misleading

hinell: Well I got no time for that. May be next time.

evilebottnawi: @hinell i mean just create issue in webpack repo and all, i am from mobile and it is hard for me right now

PR #375, opened 2019-03-01
strip query string params from targetPath

egorovli:
Strip query string from targetPath, since some assets have paths in the form of [name].[ext]?[hash]. See https://github.com/zeit/next.js/issues/6481.


egorovli:
Oh, I was in such a rush to fix I didn't even notice it's already fixed 😁

PR #377, opened 2019-03-06
Do not overwrite Content-Type

### LINKED ISSUE #376: Content-Type set by other middleware gets overwritten
* Operating System: macOS 10.14.3
* Node Version: `v11.10.0`
* NPM Version: `6.8.0`
* webpack version: `4.29.6`
* webpack-dev-middleware Version: `3.6.0`

- [x] This is a **bug**

### Code

I am working with a custom middleware to serve static assets from their gzipped versions, like so:

```javascript
app.use(function (req, res, next) {
  if (req.url.endsWith('css')) {
    req.url = req.url + '.gz';
    res.set('Content-Encoding', 'gzip');
    res.set('Content-Type', 'text/css');
  }
  next()
});
```

Unfortunately, `webpack-dev-middleware` overwrites `Content-Type` regardless of whether or not it's been set already by another middleware. See

https://github.com/webpack/webpack-dev-middleware/blob/e5bd8f8d1b12543938601f1daab40e6173db7489/lib/middleware.js#L82

### Expected Behavior

In the resulting request: `Content-Type: text/css`

### Actual Behavior

In the resulting request: `Content-Type: application/gzip; charset=UTF-8`

dmohns:

When the Content-Type header is already present (for example it has been 
set by other middleware) we do not overwrite it.

Yes, added unit test to the suite.

**Summary**

Closes #376

No

**Other information**

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -79,7 +79,9 @@ module.exports = function wrapper(context) {
           contentType += '; charset=UTF-8';
         }
 
-        res.setHeader('Content-Type', contentType);
+        if (!res.getHeader('Content-Type')) {
+          res.setHeader('Content-Type', contentType);
+        }
         res.setHeader('Content-Length', content.length);
 
         const { headers } = context.options;
--- a/test/tests/server.js
+++ b/test/tests/server.js
@@ -259,6 +259,27 @@ describe('Server', () => {
     });
   });
 
+  describe('custom Content-Type', () => {
+    before((done) => {
+      app = express();
+      const compiler = webpack(webpackConfig);
+      instance = middleware(compiler, {
+        stats: 'errors-only',
+        logLevel,
+        headers: { 'Content-Type': 'application/octet-stream' }
+      });
+      app.use(instance);
+      listen = listenShorthand(done);
+    });
+    after(close);
+
+    it('Do not guess mime type if Content-Type header is found ', (done) => {
+      request(app).get('/bundle.js')
+        .expect('Content-Type', 'application/octet-stream')
+        .expect(200, done);
+    });
+  });
+
   describe('custom mimeTypes', () => {
     before((done) => {
       app = express();


lib/middleware.js
```diff
@@ -79,7 +79,9 @@ module.exports = function wrapper(context) {
           contentType += '; charset=UTF-8';
         }
 
-        res.setHeader('Content-Type', contentType);
+        if (!res.getHeader('Content-Type')) {
+          res.setHeader('Content-Type', contentType);
+        }
         res.setHeader('Content-Length', content.length);
```
evilebottnawi: What do you think should we add check here too? I think in theory it is also possible to set own `Content-Length`, don't known real use case, but still think what we should add check here too

dmohns: This was my initial thought too. But I think matters are more complicated... It looks like ExpressJS itself sets the `Content-Length` 

https://github.com/expressjs/express/blob/3d10279826f59bf68e28995ce423f7bc4d2f11cf/lib/response.js#L194

In fact line 85 seems to be not functional at all.

evilebottnawi: @dmohns good catch, can you create issue? Will be great if you help to use investigate this, let's merge this as is

dmohns: 👍 Will do.

dmohns:
Unfortunately, I just realized the test in this PR is "wrong". The headers get set *after* Content-Type already got overwritten. Hence we are not really testing the scope of this PR. I'll shoot a separate PR with a fix.

evilebottnawi:
@dmohns :+1:

lib/middleware.js
HeGangSZ: hello，this commit make error，"TypeError: res.getHeader is not a function"，my node version is v10.15.2，how can i solve this problem

HeGangSZ: > 👍 Will do.

hello，this commit make error，"TypeError: res.getHeader is not a function"，my node version is v10.15.2，how can i solve this problem

Issue #380, opened 2019-03-07
Response.getHeader may be undefined, not a function


sjy:
* Operating System: OSX
* Node Version: 10.15.1
* NPM Version: 6.4.1
* webpack version: 4.1.6
* webpack-dev-middleware Version: 3.6.1

- [x] This is a **bug**

### Expected Behavior
No error

### Actual Behavior
A TypeError throwed
```bash
 if (!res.getHeader('Content-Type')) {
...
}
```
![image](https://user-images.githubusercontent.com/1523060/53937052-e7f8aa80-40e6-11e9-85ee-207536d4aec6.png)

### For Bugs; How can we reproduce the behavior?
Use vue-cli create a app and install the latest version of webpack-dev-middleware

sjy:
https://github.com/webpack/webpack-dev-middleware/pull/377 introduce this issue

sjy:
@evilebottnawi  should i need to make a pr which will  check `getHeader` before call it

evilebottnawi:
@sjy yep, feel free to send a PR, anyway what is use case where `getHeader` is undefined?

sjy:
My project depends on `koa-webpack`, https://github.com/shellscape/koa-webpack/issues/109

evilebottnawi:
@sjy we don't support officially `koa`, we can add check, but in future it can get other broken part

dmohns:
Looks like help is on the way: https://github.com/shellscape/koa-webpack/pull/110

hovoodd:
It's a `koa-webpack` issue, not `webpack-dev-middleware`

nuintun:
Same problem !!

wssgcg1213:
same problem.

weipxiu:
无法解决的疑难问题，似乎是绝症

davinma:
在 `"vuepress": "^0.14.10"` 下面新增一行 `"webpack-dev-middleware": "^3.6.0"` 即可。刚开始用 npm / yarn 安装的时候都报这个错误，参考 [issues/1417](https://github.com/vuejs/vuepress/issues/1417) 解决了，估计是版本号不兼容。

evilebottnawi:
All comments are not in English are marked as spam.

evilebottnawi:
Update your deps

white87332:
same problem!!

evilebottnawi:
Update your deps

emilsedgh:
[...]

PR #381, opened 2019-03-07
check res.getHeader type before exec

sjy:
https://github.com/webpack/webpack-dev-middleware/issues/380

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -79,7 +79,7 @@ module.exports = function wrapper(context) {
           contentType += '; charset=UTF-8';
         }
 
-        if (!res.getHeader('Content-Type')) {
+        if (!res.getHeader || (typeof res.getHeader === 'function' && !res.getHeader('Content-Type'))) {
           res.setHeader('Content-Type', contentType);
         }
         res.setHeader('Content-Length', content.length);
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "webpack-dev-middleware",
-  "version": "3.6.1",
+  "version": "3.6.2",
   "description": "A development middleware for webpack",
   "main": "index.js",
   "files": [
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webpack-dev-middleware",
-  "version": "3.6.1",
+  "version": "3.6.2",
```
evilebottnawi: Avoid this

lib/middleware.js
```diff
@@ -79,7 +79,7 @@ module.exports = function wrapper(context) {
           contentType += '; charset=UTF-8';
         }
 
-        if (!res.getHeader('Content-Type')) {
+        if (!res.getHeader || (typeof res.getHeader === 'function' && !res.getHeader('Content-Type'))) {
```
evilebottnawi: Can we simplify this? Looks very misleading, like:
```
if (res.getHeader && !res.getHeader('Content-Type')) {
}
```

dmohns: I think the question to be answered first is: What to do in the case where `res.getHeader` is not available. Execute the `if` statement or not?

@sjy in your statement you will execute the `if` statement when `res.getHeader` is not available. Is it correct, that **set** Header function is available though?

hovoodd:
There is no need for this pull request. It's a problem with koa wrapper which does not provide a `getHeader` method.

https://github.com/shellscape/koa-webpack/pull/110

lib/middleware.js
sjy: since the use case only effects  project using the latest webpack-koa which will be fixed, this pr is will be closed.  @dmohns, thanks your attention, code changes here is more like a hot-fix.

sjy: since the use case only effects  project using the latest webpack-koa which will be fixed, this pr is will be closed.  @dmohns, thanks your attention, code changes here is more like a hot-fix.

PR #382, opened 2019-03-07
Let ExpressJS handle Content-Length header

### LINKED ISSUE #379: Content-Length is overwritten by ExpressJS
* Operating System: macOS 
* Node Version: `v11.10.0`
* NPM Version: `6.8.0`
* webpack version: `4.29.6`
* webpack-dev-middleware Version: `3.6.1`

- [x] This is a **bug**

### Code

Follow up to discovery in https://github.com/webpack/webpack-dev-middleware/pull/377#discussion_r263004819

It looks like the computation of `Content-Length` is not functional.

https://github.com/webpack/webpack-dev-middleware/blob/dba5e02876bb76d2301c93d20ec90f558147d5b4/lib/middleware.js#L85

ExpressJS takes care of setting `Content-Length` appropriately. See also: https://github.com/expressjs/express/issues/2893#issuecomment-185839759

In fact, even when removing the line in doubt the following test will succeed

```js
describe('Check Content-Length', () => {
  before((done) => {
    app = express();
    instance = middleware(compiler, {
      stats: 'errors-only',
      logLevel
    });
    app.use(instance);
    listen = listenShorthand(done);
  });
  after(close);

  it('Check ExpressJS Content-Length computation', (done) => {
    request(app).get('/bundle.js')
      .expect('Content-Length', '4631')
      .expect(200, done);
  });
});
```

### Expected Behavior

The code suggests that `webpack-dev-middleware` is responsible for setting `Content-Length` header.

### Actual Behavior

ExpressJS is setting `Content-Length` header.

### Discussion

From my understanding it is perfectly fine to let ExpressJS deal with `Content-Length`  header and simply remove the code from `webpack-dev-middleware`.

dmohns:

Remove obsolete code.

No, existing tests cover the change.

**Summary**

Setting the `Content-Length` header from within `webpack-dev-middleware` 
does not have any effect as it is overwriten later by ExpressJS in 
https://github.com/expressjs/express/blob/master/lib/response.js#L178-L195
Remove the obsolete code.

Closes #379 

No.

**Other information**

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -82,7 +82,8 @@ module.exports = function wrapper(context) {
         if (!res.getHeader('Content-Type')) {
           res.setHeader('Content-Type', contentType);
         }
-        res.setHeader('Content-Length', content.length);
+        // The Content-Length header is set by ExpressJS appropriately. See
+        // https://github.com/expressjs/express/blob/master/lib/response.js#L178-L195
 
         const { headers } = context.options;
         if (headers) {


dmohns:
Closes #379

dmohns:
Is there any way we can fix the CodeCov test failing?

evilebottnawi:
@dmohns don't look on code codecov, all fine

dmohns:
@evilebottnawi how would you imagine such a check like? E.g. check if we are indeed using Express and only if so, don't set the Content-Length Header?

evilebottnawi:
@dmohns maybe just check `if (res.getHeader('Content-Length'))`?

dmohns:
From my understanding Express sets the header later (if I understand correctly, right before the transport happens). So, at this point in time header will be unset and there is not much to be checked.

evilebottnawi:
@dmohns hm, looks like we need major release for this change, because in theory we can break some app

dmohns:
@evilebottnawi Ok, I agree, this change might have unforeseen consequences. We can leave this PR open for a while until you want to make a new major release.

evilebottnawi:
Thanks for PR, already merged https://github.com/webpack/webpack-dev-middleware/pull/497

PR #393, opened 2019-05-11
fix(middleware): add HEAD method support

EslamHiko:
bugfix

yes

**Summary**
fixes: https://github.com/webpack/webpack-dev-server/issues/1833

no

**Other information**
when I run tests I get lint errors : 
`Expected linebreaks to be 'LF' but found 'CRLF'` and  `Delete '␍'`
and when I disable lint and run tests only requests' tests fail in expecting `Content-Length` and `Content-Range`
I'm using windows and I tried this on Ubuntu and it's the same. so is it normal?

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ _Note: The `publicPath` property is required, whereas all other options are opti
 ### methods
 
 Type: `Array`  
-Default: `[ 'GET' ,'HEAD']`
+Default: `undefined`
 
 This property allows a user to pass the list of HTTP request methods accepted by the server.
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,9 @@ _Note: The `publicPath` property is required, whereas all other options are opti
 Type: `Array`  
 Default: `undefined`
 
-This property allows a user to pass the list of HTTP request methods accepted by the server.
+This property allows a user to set HTTP request methods accepted by the server.
+
+By default all HTTP request methods are accepted by the server.
 
 ### headers
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ _Note: The `publicPath` property is required, whereas all other options are opti
 ### methods
 
 Type: `Array`  
-Default: `[ 'GET' ]`
+Default: `[ 'GET' ,'HEAD']`
 
 This property allows a user to pass the list of HTTP request methods accepted by the server.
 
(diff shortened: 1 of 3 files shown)


lib/middleware.js
```diff
@@ -44,7 +44,7 @@ module.exports = function wrapper(context) {
       });
     }
 
-    const acceptedMethods = context.options.methods || ['GET'];
+    const acceptedMethods = context.options.methods || ['GET', 'HEAD'];
```
evilebottnawi: What about support all method by defaults?

evilebottnawi:
Let's wait CI green and release as new version

Honestly, I don’t know why before that we only allowed `get` :confused:

EslamHiko:
@evilebottnawi I think because the most used method of requests is `get` so it was set to be the default. the problem was allowing `get` only.

README.md
```diff
@@ -69,7 +69,7 @@ _Note: The `publicPath` property is required, whereas all other options are opti
 ### methods
 
 Type: `Array`  
-Default: `[ 'GET' ]`
+Default: `undefined`
```
hiroppy: I think this change seems to be breaking change. Is this wrong?

hiroppy:
> Honestly, I don’t know why before that we only allowed get

Yeah, me too. So I think this change is good.

README.md
EslamHiko: @hiroppy I don't think so here is the check : 
```
const acceptedMethods = context.options.methods;

    if (acceptedMethods && acceptedMethods.indexOf(req.method) === -1) {
      return goNext();
    }
```
and in tests I changed the `post` request test that supposed to `return 404` in the previous settings to be like `get` now it passes the test without issues. 

I changed the documentation a little in the last commit, so now it makes more sense.

evilebottnawi: Yes, it is not breaking change, it is feature

evilebottnawi:
After investigate in other middleware code (one of popular exmaples https://github.com/expressjs/serve-static/blob/master/index.js#L73) i find what i was wrong, we should handle only `GET` and `HEAD` by default

evilebottnawi:
I send a PR, sorry for this

evilebottnawi:
Close in favor https://github.com/webpack/webpack-dev-middleware/pull/398, sorry again for misleading

Issue #312, opened 2018-07-06
Support webpack-cli options and/or webpack multiple configuration export


DBosley:
> The Issues page for this repository is not a support forum. **Please ask questions on StackOverflow or the webpack Gitter (https://gitter.im/webpack/webpack). Questions will be closed.** If you proceed with this form, please fill out **_all_** fields, or your issue may be closed as "invalid." Please do not delete this template from a new issue. Please remove this header to acknowledge this message.

Operating System: Ubuntu 18.04
Node Version: 8.11.1
NPM Version: 6.0.0
webpack Version: 4.12.0
webpack-hot-client Version: 4.0.3

- [x] This is a **feature** request
- [ ] This is a **bug**

### What is the motivation and/or use-case for the feature?
Webpack supports multiple configurations in a single file by exporting an array.
[Info Here](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations)
A new feature that was added recently (but currently lacks total documentation) is that you can set a 'name' property in your webpack config and pass `--config-name` to the cli to choose a specific config to build if you don't want to build all of them.

This feature request is to allow support of this functionality in the webpack-dev-middleware and related plugins.

Implementing this could be done in a couple ways:
- Support passing webpack-cli options into the middleware somehow
- Support choosing a config-name in the middleware options

I think the first solution is going to be the best approach if it's possible. This is because it would allow any webpack-cli arguments to be used with middlewares. If for some reason there's no way to go down this route, the second option would suffice for my use case, but it might be a good idea to look at other options listed [here](https://webpack.js.org/api/cli/) to see if they would merit being included as well

shellscape:
Thanks for opening an issue :beer:

> Support passing webpack-cli options into the middleware somehow

`webpack-cli` is a mess of ad-hoc options added over time, so this would not be preferable. `webpack-command` is a vastly superior CLI, but even supporting all of the flags/options available on that would be bananas.

> Support choosing a config-name in the middleware options

This would actually be much more preferable, as it's a targeted, specific use option. The important distinction here between a module like this and something like a CLI is that this module is intended to be used as API directly. We should only ever add only the options we need.

Overall I think this is a really reasonable feature request. I haven't looked into how complex the change would be, but it should be doable.

DBosley:
[...]

PR #440, opened 2019-07-08
feat(middleware): add cli options support

EslamHiko:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
solve: https://github.com/webpack/webpack-dev-middleware/issues/312

### Breaking Changes
No

### Additional Info
The unsupported Options: fs,stats,reporter,logger,mimeTypes.


EslamHiko:
@evilebottnawi did I understood the issue here https://github.com/webpack/webpack-dev-middleware/issues/312 ?

hiroppy:
I agree with @evilebottnawi.

EslamHiko:
@hiroppy okay, I think I misunderstood what the issue was about. as I understand now we can create a config file contains different sets of settings called ex `webpack.middleware.js` and then we can use 2 options ex: `configPath` and `mode` it'll import the settings in the config file with the wanted mode name? so do I understand it clearly now?

hiroppy:
I didn't join this talk but I don' think CLI is necessary. What is the use case for this feature?

evilebottnawi:
No need this here, thanks for the PR

PR #459, opened 2019-08-29
ci: webpack@5 test

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

test on webpack@5

### Breaking Changes

No

### Additional Info

No

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -12,63 +12,68 @@ const DevMiddlewareError = require('./DevMiddlewareError');
 module.exports = {
   toDisk(context) {
     const compilers = context.compiler.compilers || [context.compiler];
+
     for (const compiler of compilers) {
-      compiler.hooks.afterEmit.tap('WebpackDevMiddleware', (compilation) => {
-        const { assets } = compilation;
-        const { log } = context;
-        const { writeToDisk: filter } = context.options;
-        let { outputPath } = compiler;
+      compiler.hooks.emit.tap('WebpackDevMiddleware', (compilation) => {
+        compiler.hooks.assetEmitted.tapAsync(
+          'WebpackDevMiddleware',
+          (file, content, callback) => {
+            let targetFile = file;
 
-        if (outputPath === '/') {
-          outputPath = compiler.context;
-        }
+            const queryStringIdx = targetFile.indexOf('?');
 
-        for (const assetPath of Object.keys(assets)) {
-          let targetFile = assetPath;
+            if (queryStringIdx >= 0) {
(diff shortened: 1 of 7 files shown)


lib/fs.js
```diff
@@ -12,63 +12,68 @@ const DevMiddlewareError = require('./DevMiddlewareError');
...
+              targetFile = targetFile.substr(0, queryStringIdx);
+            }
 
-          const queryStringIdx = targetFile.indexOf('?');
+            let { outputPath } = compiler;
 
-          if (queryStringIdx >= 0) {
-            targetFile = targetFile.substr(0, queryStringIdx);
-          }
+            // TODO Why? Need remove in future major release
```
hiroppy: Can we remove this when we release the next version?

lib/fs.js
```diff
@@ -12,63 +12,68 @@ const DevMiddlewareError = require('./DevMiddlewareError');
...
+            const { log } = context;
+            const dir = path.dirname(targetPath);
+
+            return mkdirp(dir, (mkdirpError) => {
+              if (mkdirpError) {
+                return callback(mkdirpError);
+              }
+
+              return fs.writeFile(targetPath, content, (writeFileError) => {
+                if (writeFileError) {
```
hiroppy: Should we output `Unable to write asset to disk: xxx` as an error?

lib/fs.js
evilebottnawi: Yes, i think we should drop this in next major, i keep this because i can't understand why it is here :confused: Any ideas?

lib/fs.js
evilebottnawi: I think it is unnecessary, we failed compilation if something not written as expected, because it is potentially break appication

lib/fs.js
hiroppy: Yes, it's a good idea which is keeping. I don't know why... 
We'll drop this at other pr.

lib/fs.js
hiroppy: ok, agree with you.

redbugz:
This seems to have broken our dev server. We have `writeToFile` set to true, but with 3.7.1 we no longer get any files written.
I added some logging statements, and the `compiler.hooks.emit.tap` callback is called, but the `compiler.hooks.assetEmitted.tapAsync` callback never seems to get called.

PR #478, opened 2019-09-30
refactor: utils

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Second part of refactor (i think we will do this for 3-4 parts)

### Breaking Changes

No

### Additional Info

No

/cc @hiroppy


lib/utils/setupWriteToDisk.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const { colors } = require('webpack-log');
+
```
hiroppy: Why do we need a new line at here?

evilebottnawi: @hiroppy fixed

hiroppy:
Thanks!

PR #481, opened 2019-10-02
refactor: utils

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Simplify code

### Breaking Changes

No

### Additional Info

No


lib/middleware.js
```diff
@@ -131,6 +127,27 @@ module.exports = function wrapper(context) {
...
+            processRequest();
+
+            return;
+          }
+        } catch (_error) {
+          // Ignore error
+        }
+      }
+
+      ready(context, processRequest, req);
```
evilebottnawi: Move this inside file, no need util for simple action, also it is hard to read, because you should look in two files when request happens and it is painful

lib/middleware.js
```diff
@@ -131,6 +127,27 @@ module.exports = function wrapper(context) {
...
+
+      if (HASH_REGEXP.test(filename)) {
+        try {
+          if (context.fs.statSync(filename).isFile()) {
+            processRequest();
+
+            return;
+          }
+        } catch (_error) {
+          // Ignore error
```
hiroppy: Is this ok if we ignore this error?

evilebottnawi: Yes, we should check `html` index file if we can't find bundle, so we need ignore this error, maybe we can change logic in future version

PR #482, opened 2019-10-04
refactor: code

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

- update webpack-default, 
- move `index.js` inside lib (let's do not mix configuration files in root with code)
- forgot `'use strict'` in some places

### Breaking Changes

No

### Additional Info

No


evilebottnawi:
After this PR i will start rewriting logger on webpack logger

.eslintignore
```diff
@@ -0,0 +1,4 @@
+/coverage
+/dist
+/node_modules
+/test/fixtures
```
hiroppy: We need the line break?

.prettierignore
```diff
@@ -1,3 +1,5 @@
-coverage
-node_modules
-CHANGELOG.md
+/coverage
+/dist
+/node_modules
+/test/fixtures
+CHANGELOG.md
```
hiroppy: ditto

package.json
```diff
@@ -7,25 +7,26 @@
...
+    "lint:js": "eslint --cache lib test",
+    "lint": "npm-run-all -l -p \"lint:**\"",
     "release": "standard-version",
     "security": "npm audit",
-    "test:only": "jest",
-    "test:coverage": "npm run test:only -- --coverage",
-    "test:watch": "npm run test:only --watch",
+    "test:only": "cross-env NODE_ENV=test jest",
+    "test:watch": "cross-env NODE_ENV=test jest --watch",
+    "test:coverage": "cross-env NODE_ENV=test jest --collectCoverageFrom=\"lib/**/*.js\" --coverage",
```
hiroppy: It is better to separate to jest.config.js?

evilebottnawi: I think it is more simple than using `jest.config.js`. Maybe we revisit this in future.

.eslintignore
evilebottnawi: fixed

.prettierignore
evilebottnawi: fixed

package.json
hiroppy: ok

PR #520, opened 2019-11-20
refactor: remove the `lazy` option

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Remove the `lazy` option

### Breaking Changes

Yes

### Additional Info

This option is not compatible with webpack behavior.


src/options.json
```diff
@@ -46,12 +46,6 @@
     "headers": {
       "type": "object"
     },
-    "lazy": {
-      "type": "boolean"
-    },
-    "filename": {
```
hiroppy: filename is also unnecessary?

evilebottnawi: Yes, filename only for `lazy` mode

hiroppy: understood, thanks

PR #567, opened 2020-02-04
Refactor code

evilebottnawi:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

We never officially support `koa` and this is not mentioned anywhere in the documentation so I rewrite code on official express api and remove unnecessary test

### Breaking Changes

In theory yes, but we didn’t indicate it anywhere in the documentation and this is a major release, we can refactor

### Additional Info

Maybe we should consider supporting `koa`, but we need time for that and more tests. Therefore, if anyone wants to introduce `koa` support, let it do it fully. We are open to improvements.


evilebottnawi:
Feel free to feedback

PR #576, opened 2020-02-10
Refactor get filename from url

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Changes:
- respect `publicPath` value from `output.path`
- fix using `[hash]`/`[fullhash]` in `output.path` and `output.publicPath` options
- fix broken test
- remove invalid check on `HASH`, a developer can use any filename for any files
- we need use `stats` to get output path and public path

### Breaking Changes

Yes, default value of `publicPath` is `output.publicPath`

### Additional Info

We recommend use a same value for `publicPath` and for `output.publicPath` so let's do it by default for 0CJS https://github.com/webpack/webpack-dev-middleware#publicpath

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,6 @@ Some of the benefits of using this middleware include:
   has completed.
 - Supports hot module reload (HMR).
 
-## Requirements
-
-This module requires a minimum of Node v6.9.0 and Webpack v4.0.0, and must be used with a
-server that accepts express-style middleware.
-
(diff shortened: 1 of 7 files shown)


evilebottnawi:
Feel free to leave feedback

PR #578, opened 2020-02-11
Refactor code

evilebottnawi:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

There was a lot of wrong logic:
- `request.url` is always `/path/to/file.js?foo=bar#hash`, so it is unnecessary to check on `host`
- we should avoid check on the `host` because it is middleware, we don't know what hosts a developer what to use, it should be done on application side (we already do it in webpack-dev-server)
- memorize the `parse`, no need parse multiple times same urls
- remove invalid tests
- refactor tests
- new tests
- when you use multi-compiler mode you can use same "publicPath" but difference "output.path" and versa vise, so we need return array of possible files, also some a child compiler can not have "publicPath"
- rename to `getPossibleFilePaths`

### Breaking Changes

Yes, there are a lot of fixes + remove `getPossibleFilePaths` from API, it is doesn't have sense to do public, because we can't use it before compilation, after compilation you can get paths from stats

### Additional Info

No

### DIFF
--- a/src/utils/getFilenameFromUrl.js
+++ b/src/utils/getFilenameFromUrl.js
@@ -54,6 +54,7 @@ export default function getFilenameFromUrl(context, url, stats) {
   const { options } = context;
   const { outputPath, publicPath } = getPaths(stats, options, url);
   // localPrefix is the folder our bundle should be in
+  // TODO catch error
   const localPrefix = parse(publicPath || '/', false, true);
   const urlObject = parse(url);
   const hostNameIsTheSame = localPrefix.hostname === urlObject.hostname;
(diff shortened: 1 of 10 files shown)


src/utils/getFilenameFromUrl.js
```diff
@@ -1,117 +1,75 @@
...
-    return false;
+  try {
+    // The `url` property of the `request` is contains only  `pathname`, `search` and `hash`
+    urlObject = memoizedParse(url, false, true);
+  } catch (_ignoreError) {
+    return pathToFile;
   }
 
-  let uri = outputPath;
+  for (const childPaths of paths) {
```
jamesgeorge007:
```
  for (const { publicPath, outputPath } of paths) {
```

jamesgeorge007: What about this approach?

evilebottnawi: Already fixed :+1:

test/utils/getPossibleFilePaths.test.js
```diff
@@ -1,12 +1,18 @@
-import getFilenameFromUrl from '../../src/utils/getFilenameFromUrl';
+import getFilenameFromUrl from '../../src/utils/getPossibleFilePaths';
```
evilebottnawi: I will rewrite these tests to real tests in the next PR to ensure all works fine with express too

Issue #602, opened 2020-03-19
4.0.0-rc.1 is incompatible with Fastify


artemirq:
- Node Version: v13.11.0
- NPM Version: 6.13.7
- webpack Version: 4.42.0
- webpack-dev-middleware Version: 4.0.0-rc.1

### Actual Behavior

When I trying to access application from browser, build is failing with error `TypeError: res.get is not a function`. In previous version, 3.7.2 it's not happened

I checked that error is happening in new version on [this code line](https://github.com/webpack/webpack-dev-middleware/blob/master/src/middleware.js#L59)

### Code

```js
const webpack = require('webpack');
const middleware = require('webpack-dev-middleware').default;
const path = require('path');
const compiler = webpack({
  entry: './src/index.js',
  output: {
    filename: 'main.js',
    path: path.resolve(__dirname, 'dist'),
  },
});
const fastify = require('fastify');
const app = fastify();

app.use(
  middleware(compiler)
);

app.listen(3000, () => console.log('Example app listening on port 3000!'));
```

### How Do We Reproduce?

1. Bootstrap project from scratch with `npm init`
2. Run `npm i webpack webpack-dev-middleware@next fastify`
3. Create file `src/index.js` and write something valid
4. Place the code above to `./run-webpack.js`
5. Run `node run-webpack.js`
6. Try to access http://localhost:3000/main.js in your browser

### Notes

- I also checked that my example works with Express normally, but with Fastify not

evilebottnawi:
We never will be compatibility with `fastify`, we support only `express` officialy

artemirq:
Supporting availability to use with two frameworks is bad idea, agreed

evilebottnawi:
In theory we can implement it, but we need improve tests, let's mark it is feature

pearofducks:
@artemirq (and any others who find this) - this is working fine with the lastest fastify and [fastify-express](https://github.com/fastify/fastify-express) as the middleware provider (instead of middie)

evilebottnawi:
@pearofducks yep, can you send a PR to docs how to use it with fastify?

WraithKenny:
This is also breaking with `browser-sync` where before it was working fine. (In my setup, I have a Gulp file which runs a BrowserSync server, which loads the middleware.) It's the same issue here; should I make a new issue, or just leave this note?

evilebottnawi:
we never supports faslity, please look at README, we write how you should use it with fastify

[...]

PR #617, opened 2020-04-09
feat: add support for more frameworks other than express only

EslamHiko:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
fix: https://github.com/webpack/webpack-dev-middleware/issues/602

### Breaking Changes
No

### Additional Info
I'm not sure if my tests are enough.


test/middleware.test.js
```diff
@@ -2663,7 +2864,30 @@ describe('middleware', () => {
...
+    afterAll(close);
+
+    it('should return the "200" code for the "GET" request to the bundle file and return headers', (done) => {
+      request(app.server)
+        .get('/bundle.js')
+        .expect('X-nonsense-1', 'yes')
+        .expect('X-nonsense-2', 'no')
+        .expect(200, done);
+    });
+  });
```
evilebottnawi, line 2890: Please rewrite it:
```
const framewroks = [express, fastify];

for (const framework of framewroks) {
  // Tests
}
```

src/middleware.js
```diff
@@ -69,13 +69,16 @@ export default function wrapper(context) {
...
           }
         }
 
-        res.send(content);
-
+        if (res.send) {
+          res.send(content);
+        } else {
+          res.end(content);
+        }
```
evilebottnawi, line 81: Need encapsulate to utils with comments - `getHeader`, `setHeader`, `send` with code like:
```js
function getHeader(res, header) {
  // Fastify
  if (res.getHeader) {
    return res.getHeader(header)
  }

  // Express
  return res.get(header)
}
```

Maybe fastify has something in res what allow to known what it is fastify

EslamHiko: @evilebottnawi I think `getHeader` and `setHeader` works with all frameworks as `res` extend [responseServer](https://nodejs.org/dist/latest/docs/api/http.html#http_class_http_serverresponse)
the difference is that `.send` is express only function ref: https://github.com/fastify/fastify/issues/319

test/middleware.test.js
EslamHiko: @evilebottnawi as fastify don't support `.send` the header `content-length` will be missing from all the tests.

EslamHiko:
Some Important notes about my last commit : 
- for [`404` and errors](https://github.com/webpack/webpack-dev-middleware/blob/master/src/middleware.js#L54) express `Content-Type` header is always `text/html; charset=utf-8` but fastify `Content-Type` is `application/json; charset=utf-8`, I tried to set the header before `resolve(goNext())` but it threw an error.
- express `send` provides `Content-Length` and `Content-Type` headers and  so I created [src/utils/getContentLength.js](https://github.com/webpack/webpack-dev-middleware/pull/617/commits/fb4a63f144fae030ad9b1de17cf3f56dbba5d179#diff-056a7a9e073d7a0e42508732b74c0292) and [src/utils/getContentLength.js](https://github.com/webpack/webpack-dev-middleware/pull/617/commits/fb4a63f144fae030ad9b1de17cf3f56dbba5d179#diff-821ba371dcb5ff7fd6133265d76b7209) so I can add these headers manually before `res.end()` I took the logic from [express.send](https://github.com/expressjs/express/blob/master/lib/response.js#L107)
- I changed `phtml` [here](https://github.com/webpack/webpack-dev-middleware/pull/617/commits/fb4a63f144fae030ad9b1de17cf3f56dbba5d179#diff-c812c858fe33f1b87cdeb92e2bdc8567R2113) to `phtml2` and `custom` [here](https://github.com/webpack/webpack-dev-middleware/pull/617/commits/fb4a63f144fae030ad9b1de17cf3f56dbba5d179#diff-c812c858fe33f1b87cdeb92e2bdc8567R3168) to `customext` because when I run all the tests they don't pass however If I run both of them a lone they pass, that happened because `mimeTypes['custom']` & `mimeTypes['phtml']` existed maybe because the tests are sync or the server didn't close proberly or there was some kind of cache, however changing them doesn't break the logic of the tests because `phtml2` & `customext` still are unknown MIME types.

src/utils/getContentLength.js
```diff
@@ -0,0 +1,18 @@
+export default function getContentLength(content) {
+  let len;
```
hiroppy, line 2: Don't use undefined as an initial value.  
Need to validate here(https://github.com/webpack/webpack-dev-middleware/pull/617/files#diff-f74cf7ea3fda676a7bafaf4bd8714ea4R83) or set 0.

evilebottnawi:
I think we need to add support for:
- fastify
- koa
- hapi

EslamHiko:
@evilebottnawi sure, I'll add the other frameworks in the test as well :smiley:

evilebottnawi:
@EslamHiko great! Thanks for help

EslamHiko:
@evilebottnawi I'm afraid I can't support koa and hapi both are totally different than express. 

Hapi must have plugins like this: https://hapi.dev/tutorials/expresstohapi/?lang=en_US#creating
Koa doesn't pass the traditional (req, res, next).

Now it supports all the framework that its response object extends Node.js [http.ServerResponse](https://nodejs.org/docs/v0.4.0/api/http.html#http.ServerResponse)

evilebottnawi:
I think we need test `polka`.

- Hapi - I think we can create helper:
```
const webpackDevMiddleware = require('webpack-dev-middleware').koa;
```

What do you think?

- Koa - can you clarify? I think we can rewrite some of parts to support it

src/middleware.js
```diff
@@ -56,26 +57,26 @@ export default function wrapper(context) {
         return;
       }
 
-      if (!res.get('Content-Type')) {
+      if (!res.getHeader('Content-Type')) {
```
evilebottnawi, line 60: I think we can create `compatibility.js` file and put all helpers in one file with comments:
```
// `res.get` used by express
// `res.getHeader` used by fastify
// etc
```

hiroppy:
Let's separate PR if you support others. I think the minimum start is better and to discuss this topic is good but I want to discuss at issues.

src/middleware.js
EslamHiko: @evilebottnawi I think it'll be better to have getHeader, setHeader & end as the default functions as they're the default for Node.js http.ServerResponse and write `handleCompatibility.js`: 
```js
export default function handleCompatibility(ctx,res) {
  if(typeof res.getHeader === 'undefined'){
    res.getHeader = res.get
  }
  if(typeof res.setHeader === 'undefined'){
    res.setHeader = res.set
  }
  // ctx is koa context
  if(typeof res.end === 'undefined' && ctx){
    res.end = (content) => ctx.body = content;
  }
  res.setStatus = (code) => {
    res.statusCode = code;
    res.status = code;
  }
}
```
Note: Ctx is for koa js.

EslamHiko:
@hiroppy I totally agree, I could work on supporting koa and I passed most of the tests but it has problems with changing `publicPath` It'll need so much debugging. So let's review this PR.

src/middleware.js
evilebottnawi: Modified `res` and `req` is very unsafe, we should avoid it

src/middleware.js
evilebottnawi: @EslamHiko hm, if is is supported why we change it?

evilebottnawi:
@hiroppy 

> Let's separate PR if you support others.

Agree

> I think the minimum start is better and to discuss this topic is good but I want to discuss at issues.

We can discuss this. 

My plan is pretty simple:
- extract webpack hot client from `webpack-dev-server` (it will be done for v5, I will create roadmap) to own middleware and deprecated https://github.com/webpack-contrib/webpack-hot-client and https://github.com/webpack-contrib/webpack-hot-middleware (maybe we will get `webpack-hot-middleware` name)
- refactor webpack-dev-server using `webpack-hot-middleware`
- improve compatibility with other server frameworks (not only express)

Pros:
- Developers will be able to configure own server as they need without any hacks or using outdated packages
- Official maintainable hot middleware
- Easy to tests dev server and easy to maintenance
- Ability to create own dev servers (it would be very convenient for complex solutions) and ability to build a server using any popular server framework
- happy faces

I believe that we can start this work here by introducing compatibility with popular solutions, this list may increase in the future, but I do not see any problems with testing.

src/middleware.js
EslamHiko: I was wrong about `getHeader and setHeader works with all frameworks` I thought they all extend `ServerResponse`, what do you mean by `is`?

hiroppy:
@evilebottnawi yes, agree with your Pros. @EslamHiko can you work on this plan?

EslamHiko:
@hiroppy I can work on `improve compatibility with other server frameworks` I need more info for `extract webpack hot client from webpack-dev-server` how can I extract it?, about this PR is there anything else I can do?

src/middleware.js
EslamHiko: @evilebottnawi for koa I have to modify them because koa passes ctx and next for ex:
```js

app.use(async (ctx, next) => {
  await next();
  const rt = ctx.response.get('X-Response-Time');
  console.log(`${ctx.method} ${ctx.url} - ${rt}`);
});
```
What I did was to do this:
```js
  return async function middleware(...args) {
    let ctx;
    let req;
    let res;
    let next;
    // for detecting koa framework
    if(typeof args[1] === "function"){
      ctx = args[0];
      next = args[1]
      req = ctx.request
      res = ctx.response
    } else {
      req = args[0]
      res = args[1]
      next = args[2]
    }
   ......
```

evilebottnawi:
@EslamHiko 

>  I need more info for extract webpack hot client from webpack-dev-server how can I extract it

We'll come back to this after implementing support for most of the solutions here. I can even give a roadmap where you can help us

src/middleware.js
evilebottnawi: Let's focus on `fastify`, we will improve koa supporting in separate PR

I suggest to create `compatibility.js` as written above. For koa we will create wrapper around function to achieve this.

Here let's create `compatibility.js` file with two methonds:
```
function getResponseHeader() { }
function sendResponse() { }
```

src/utils/compatibility.js
```diff
@@ -0,0 +1,14 @@
...
+  return res.getHeader(header);
+};
+const resSetHeader = (res, header, value) => {
+  if (res.set) {
+    return res.set(header, value);
+  }
+  return res.setHeader(header, value);
+};
+
+export default { resGetHeader, resSetHeader };
```
evilebottnawi, line 14: Use:
```
function resGetHeader() { }
function resSetHeader() { }
```

no need to keep it in variable

evilebottnawi (commented):
Put all helpers in `compatibility.js` file, it will be easy to maintenance

EslamHiko:
@evilebottnawi currently I put `resGetHeader,resSetHeader,getContentLength,getUnknownContentType,sendContent` in `compatibility.js`.
`getFilenameFromUrl, getPaths, handleRangeHeaders, ready, setupHooks, setupOutputFileSystem, setupWriteToDisk` are separated. should I add any of them to `compatibility.js` ?

evilebottnawi:
@EslamHiko No, only utils for `compatibility.js`

evilebottnawi:
> currently I put resGetHeader,resSetHeader,getContentLength,getUnknownContentType,sendContent in compatibility.js.

I think you forgot to send commit, also please fix lint problems `npm run lint`

src/utils/compatibility.js
```diff
@@ -1,14 +1,54 @@
-const resGetHeader = (res, header) => {
+import mime from 'mime-types';
+
+export function resGetHeader(res, header) {
   if (res.get) {
```
evilebottnawi, line 4: Put comment:
```
// For `express`
```

src/utils/compatibility.js
```diff
@@ -1,14 +1,54 @@
...
     return res.get(header);
   }
   return res.getHeader(header);
```
evilebottnawi, line 7: Put comment:
```
// For `fastify`
```

package.json
```diff
@@ -62,6 +62,7 @@
     "eslint-plugin-import": "^2.22.0",
     "eslint-plugin-prettier": "^3.1.4",
     "express": "^4.17.1",
+    "fastify": "^2.13.0",
```
evilebottnawi, line 65: Latest version is `3.3.0` https://github.com/fastify/fastify/blob/master/package.json#L3, we don't need support old versions

test/middleware.test.js
```diff
@@ -1635,7 +1710,11 @@ for (const frameworkObject of frameworks) {
           instance = middleware(compiler);
 
           app = frameworkObject.framework();
-          app.use(instance);
+          if (frameworkObject.name === 'fastify') {
+            app.register(middie).then(() => app.use(instance));
+          } else {
+            app.use(instance);
+          }
```
evilebottnawi, line 1717: Let avoid duplication

EslamHiko:
@evilebottnawi not sure why tests are hanging :confused:

src/utils/compatibility.js
```diff
@@ -0,0 +1,64 @@
...
+  const contentLength = getContentLength(content);
+  resSetHeader(res, 'Content-Length', contentLength);
+
+  if (!resGetHeader(res, 'Content-Type')) {
+    const unknownContentType = getUnknownContentType(content);
+    resSetHeader(res, 'Content-Type', unknownContentType);
+  }
+
+  return res.end(content);
+}
```
evilebottnawi, line 64: We really need `sendContent` https://www.fastify.io/docs/v1.14.x/Reply/#senddata?

EslamHiko: @evilebottnawi Yep we need it: https://github.com/fastify/fastify/issues/319

evilebottnawi: I am afraid it is not safe to do this https://github.com/fastify/fastify/pull/328/files#diff-f5fc3aad6335ea7ddc5204ffa960e59eR34

EslamHiko: @evilebottnawi `If you need to create a middleware you have to use the Node `req` and `res` instances` that's exactly what we do we use `.end` and `.setHeader` that are associated with Node `res`.

evilebottnawi: Look at example of middleware https://github.com/fastify/fastify-static/blob/master/index.js#L93, I still don't undestand why we send `Content-Type` here, can you provide more information?

evilebottnawi: Looks `reply.send` is available here

EslamHiko: we send `content-type` to pass the tests as express set the `content-type` by default, whenever I use `res.send` alone it gives `send is not a function` error.

evilebottnawi: @EslamHiko so `fastify` do not send `content-type` by default? It is strange because in other middleware I don't find code for `content-type`

EslamHiko: @evilebottnawi oh sorry `fastify` don't send `content-length` header but it sends `Content-Type: text/html; charset=utf-8` for unknown content-type files instead of `Content-Type: application/octet-stream` like express.

evilebottnawi: hm, what is Node.js (using built-in `http`/`https` module) default behavior, can you investigate?

EslamHiko: @evilebottnawi I just made a very simple Node.js server and it didn't provide the content type for unknown mime type files.
`server.js`
```js
const http = require('http');

const hostname = '127.0.0.1';
const port = 3000;
const fs = require('fs');
const file = fs.readFileSync('./file');
const server = http.createServer((req, res) => {
  res.end(file);
});

server.listen(port, hostname, () => {
  console.log(`Server running at http://${hostname}:${port}/`);
});
```
`file`
```
this is supposed to be unknown mime type file
```
headers

We can see here what express do for content types here : https://github.com/expressjs/express/blob/0a48e18056865364b2461b2ece7ccb2d1075d3c9/lib/response.js#L141
and that's what fastify do : 
https://github.com/fastify/fastify/blob/b9fcf3fae0317a54f4f32c5121d5099d1a4d0ed9/lib/reply.js#L131

EslamHiko: Some notes : 
- If I remove the content length setting section it fails in this test "should return the "200" code for the "HEAD" request to the bundle file"
```js
it('should return the "200" code for the "HEAD" request to the bundle file', (done) => {
          request(server)
            .head('/bundle.js')
            .expect(
              'Content-Length',
              instance.context.outputFileSystem
                .readFileSync(path.resolve(outputPath, 'bundle.js'))
                .byteLength.toString()
            )
            .expect('Content-Type', 'application/javascript; charset=utf-8')
            // eslint-disable-next-line no-undefined
            .expect(200, undefined, done);
        });
```
Screenshot from 2020-09-06 14-56-27

- I get this when I comment the unknown content-type setting section.
```
 middleware (fastify) › basic › should work › should return "200" code code for the "GET" request to the file without extension

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "filename-name-with-dots/mono-v6.x.x" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › basic › should work with difference requests › should work handle requests › should return the "200" code for the "GET" request for the "noextension" url

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › mimeTypes option › should set the correct value for "Content-Type" header to unknown MIME type › should return the "200" code for the "GET" request to "file.phtml"

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › index option › should work with "string" value with a custom extension › should return the "200" code for the "GET" request to the public path

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ...

  ● middleware (fastify) › index option › should work with "string" value without an extension › should return the "200" code for the "GET" request to the public path

    expected "Content-Type" header field

      at Test._assertHeader (node_modules/supertest/lib/test.js:237:45)
      ....

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
Test Suites: 1 failed, 10 passed, 11 total
Tests:       28 failed, 995 passed, 1023 total
Snapshots:   66 passed, 66 total
Time:        13.362 s, estimated 14 s
Ran all test suites.

```
It looks that fastify fixed the error with setting the content-length here : https://github.com/fastify/fastify/pull/2215
but it still doesn't support it for `head` requests.
It seems that in the new version fastify don't set `Content-Type` for unknown types at all.

- And that's what happens when I just return `res.send`.
note send

evilebottnawi: I think we should fix some tests, because in most cases `send` exists

EslamHiko: Yes it passes in most cases because I'm running all the tests (`getFileFromUrl.test.js`, `getPaths.test.js` ...) including `middleware.test.js` but if I test only `middleware.test.js` that actually uses the `.send` function about half of the tests pass (all express tests + 404 tests).
only mid
If I test fastify alone only `404` (144) tests pass which makes sense. 
error 404

evilebottnawi: It is very strange, because it must either completely pass or completely fall, can you look deeper? What is `res` for failed tests?

hiroppy:
@EslamHiko sorry could you rebase?

alexander-akait:
Now we have official readme how to work with fastify and suppor it with test, anyway thank you for your work

PR #670, opened 2020-06-27
fix(middleware): change how mime type option is handled

Loonride:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I think that the `mimeType` option should take priority over existing, known mime types. Otherwise, specifying a `mimeType` for something like html or js will have no effect.

### Breaking Changes

Change of mimeType option priority with existing mime types

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -49,15 +49,9 @@ export default function wrapper(context) {
         return;
       }
 
-      try {
-        content = context.outputFileSystem.readFileSync(filename);
-      } catch (_ignoreError) {
-        await goNext();
-        return;
-      }
-
-      // Buffer
-      content = handleRangeHeaders(content, req, res);
+      // Content-Type and headers need to be set before checking if
+      // the file is in the outputFileSystem, as these things should be
+      // applied to all files that are being served
 
       if (!res.get('Content-Type')) {
         // content-type name(like application/javascript; charset=utf-8) or false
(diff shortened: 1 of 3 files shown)


src/middleware.js
```diff
@@ -49,15 +49,9 @@ export default function wrapper(context) {
...
-      } catch (_ignoreError) {
-        await goNext();
-        return;
-      }
-
-      // Buffer
-      content = handleRangeHeaders(content, req, res);
+      // Content-Type and headers need to be set before checking if
+      // the file is in the outputFileSystem, as these things should be
+      // applied to all files that are being served
```
Loonride, line 54: correct me if this approach is incorrect, but it seems that headers and content types need to be applied before rejecting a file that is not in the outputFileSystem. This is because files/routes served by webpack-dev-server like `localhost:8080/main`, or files that are in the user's directory and are being served, are not in the middleware filesystem but should still have the headers and content types applied to them. That is what the previous behavior was. But let me know if this was an intentional breaking change.

evilebottnawi: Yes, looks like regression

evilebottnawi:
@Loonride should we do patch release? Or we have more regressions?

Loonride:
> @Loonride should we do patch release? Or we have more regressions?

@evilebottnawi Not yet, there still may be more that needs to be done. Still working through integrating it on the webpack-dev-server side

evilebottnawi:
:+1:

Loonride:
@evilebottnawi I think it is good now for a patch release

PR #672, opened 2020-06-30
fix(middleware): switch headers and mime types handling back after file check

Loonride:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Switch headers and mimeType handling back after it is checked whether the file is in outputFileSystem

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -49,9 +49,12 @@ export default function wrapper(context) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


evilebottnawi:
We need fix test too

PR #674, opened 2020-07-06
chore(utils): add getPaths helper as separate util

Loonride:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

The `getPaths` helper should be exported so that it can be used in `webpack-dev-server`: https://github.com/webpack/webpack-dev-server/pull/2671

Still need to add tests

### Breaking Changes

None

### DIFF
--- a/test/api.test.js
+++ b/test/api.test.js
@@ -198,12 +198,16 @@ describe('API', () => {
   });
 
   describe('context property', () => {
-    it('should contain public properties', () => {
+    it('should contain public properties', (done) => {
       expect(instance.context.state).toBeDefined();
       expect(instance.context.options).toBeDefined();
       expect(instance.context.compiler).toBeDefined();
       expect(instance.context.watching).toBeDefined();
       expect(instance.context.outputFileSystem).toBeDefined();
+
+      compiler.hooks.done.tap('wdm-test', () => {
+        done();
+      });
     });
   });
 });
--- a/test/api.test.js
+++ b/test/api.test.js
@@ -205,6 +205,8 @@ describe('API', () => {
       expect(instance.context.watching).toBeDefined();
       expect(instance.context.outputFileSystem).toBeDefined();
 
+      // the compilation needs to finish, as it will still be running
+      // after the test is done if not finished, potentially impacting other tests
       compiler.hooks.done.tap('wdm-test', () => {
         done();
       });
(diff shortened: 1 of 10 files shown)


hiroppy:
@Loonride This PR is WIP?

evilebottnawi:
@Loonride Still broken :disappointed:

test/api.test.js
```diff
@@ -198,12 +198,18 @@ describe('API', () => {
...
       expect(instance.context.state).toBeDefined();
       expect(instance.context.options).toBeDefined();
       expect(instance.context.compiler).toBeDefined();
       expect(instance.context.watching).toBeDefined();
       expect(instance.context.outputFileSystem).toBeDefined();
+
+      // the compilation needs to finish, as it will still be running
+      // after the test is done if not finished, potentially impacting other tests
+      compiler.hooks.done.tap('wdm-test', () => {
+        done();
```
Loonride, line 211: I couldn't reproduce the problem locally, but there was a problem with `api.test.js` impacting `getFilenameFromUrl.test.js` when it was being run directly after `api.test.js`, happening on CI for webpack@next only.

The problem was that the very last test of `api.test.js` was not waiting for compilation to finish, and somehow this was impacting the very first test of `getFilenameFromUrl.test.js` by preventing the middleware's done hook from being called for this very first test.

I'm not sure why this was happening, but it could be something with webpack@next saving some global data about compiler hooks that is still there when the new webpack compilation starts. Also still not sure why I couldn't reproduce locally, I tried to make everything identical to the CI environment and installed webpack@next but tests still passed.

Loonride:
/cc @evenstensberg @hiroppy 

Sorry if util tests are slightly repetitive to the `middleware` and `api` tests, but I don't like making use of a util in `webpack-dev-server` if that util isn't directly tested

If it's fine with everyone, I can add tests for more of the utils files

test/api.test.js
evilebottnawi: Very strange, if I comment this line, it is happens in CI, right?

Loonride: @evilebottnawi Yes

PR #675, opened 2020-07-09
chore(utils): add utils tests

Loonride:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Adding tests for util files

### Breaking Changes

Yes, breaking API change for handleRangeHeaders util

### DIFF
--- a/src/utils/handleRangeHeaders.js
+++ b/src/utils/handleRangeHeaders.js
@@ -1,6 +1,6 @@
 import parseRange from 'range-parser';
 
-export default function handleRangeHeaders(content, req, res) {
+export default function handleRangeHeaders(context, content, req, res) {
   // assumes express API. For other servers, need to add logic to access
   // alternative header APIs
   res.setHeader('Accept-Ranges', 'bytes');
@@ -13,11 +13,18 @@ export default function handleRangeHeaders(content, req, res) {
       res.setHeader('Content-Range', `bytes */${content.length}`);
       // eslint-disable-next-line no-param-reassign
       res.statusCode = 416;
-    }
-
-    // valid (syntactically invalid/multiple ranges are treated as a
-    // regular response)
-    if (ranges !== -2 && ranges.length === 1) {
+    } else if (ranges === -2) {
+      // malformed header treated as regular response
+      context.logger.error(
+        'A malformed Range header was provided. A regular response will be sent for this request.'
+      );
+    } else if (ranges.length !== 1) {
+      // multiple ranges treated as regular response
+      context.logger.error(
+        'A Range header with multiple ranges was provided. Multiple ranges are not supported, so a regular response will be sent for this request.'
+      );
+    } else {
+      // valid range header
       const { length } = content;
 
       // Content-Range
--- a/test/middleware.test.js
+++ b/test/middleware.test.js
@@ -229,6 +229,17 @@ describe('middleware', () => {
           .expect(206, done);
       });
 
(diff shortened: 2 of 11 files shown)


test/utils/handleRangeHeaders.test.js
```diff
@@ -0,0 +1,75 @@
...
+      },
+    };
+
+    const res = {
+      setHeader: jest.fn(),
+    };
+
+    const contentRes = handleRangeHeaders(content, req, res);
+    expect(contentRes).toEqual('abcdef');
+    expect(res.statusCode).toBeUndefined();
```
Loonride, line 36: @evilebottnawi Is this expected behavior that the statusCode is not set to something on malformed range header?

test/utils/handleRangeHeaders.test.js
```diff
@@ -0,0 +1,75 @@
...
```
Loonride, line 72: Same here

test/utils/handleRangeHeaders.test.js
evilebottnawi: We need looks in docs/specs, i think statusCode should always be

test/middleware.test.js
```diff
@@ -229,6 +229,17 @@ describe('middleware', () => {
           .expect(206, done);
       });
 
+      it('should return the "200" code for the "GET" request with malformed range header which is ignored', (done) => {
+        request(app).get('/bundle.js').set('Range', 'abc').expect(200, done);
+      });
```
Loonride, line 234: @evilebottnawi I think a 200 response code is fine since we are intentionally ignoring these header cases.

evilebottnawi: Agree

test/utils/handleRangeHeaders.test.js
evilebottnawi: I think it is ok to ignore malformed range, log error in console and return original code, i.e. `200` if no problems before

src/utils/handleRangeHeaders.js
```diff
@@ -1,6 +1,6 @@
 import parseRange from 'range-parser';
 
-export default function handleRangeHeaders(content, req, res) {
+export default function handleRangeHeaders(context, content, req, res) {
```
Loonride, line 3: Made a breaking change here because I think context should come first

src/utils/handleRangeHeaders.js
```diff
@@ -13,11 +13,18 @@ export default function handleRangeHeaders(content, req, res) {
...
-    if (ranges !== -2 && ranges.length === 1) {
+    } else if (ranges === -2) {
+      // malformed header treated as regular response
+      context.logger.error(
+        'A malformed Range header was provided. A regular response will be sent for this request.'
+      );
+    } else if (ranges.length !== 1) {
+      // multiple ranges treated as regular response
+      context.logger.error(
+        'A Range header with multiple ranges was provided. Multiple ranges are not supported, so a regular response will be sent for this request.'
```
Loonride, line 24: @evilebottnawi Logging it now as an error and proceeding with a normal response

src/utils/handleRangeHeaders.js
evilebottnawi: All is fine, we are in rc

PR #712, opened 2020-09-14
test(middleware): remove snapshots of output logs

hiroppy:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Our CI fails when we introduce webpack@next(from beta.30). 
Output log depends on webpack's version so I removed it.

see https://github.com/webpack/webpack-dev-middleware/pull/711


test/middleware.test.js
```diff
@@ -3654,7 +3654,9 @@ describe('middleware', () => {
       it('should logging', (done) => {
         compiler.hooks.failed.tap('FailedCatcher', () => {
           instance.close(() => {
-            expect(getLogsPlugin.logs).toMatchSnapshot();
+            expect(getLogsPlugin.logs[0]).toEqual(
+              expect.arrayContaining(['error'])
+            );
```
evilebottnawi, line 3659: It is bad testing :disappointed: We can have error from loader/plugin or other, we need snapshots here

hiroppy: hmm, from beta.30, snapshot gets webpack-version and some stuff which depends on webpack's version.... And how to switch snapshot between v4 and v5?

evilebottnawi: We can normalize snapshots between versions or use https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/test/helpers/snapshotResolver.js

hiroppy: I see, thanks. I'll try it.

evilebottnawi (commented):
/cc @hiroppy can we focus on it?

evilebottnawi:
Fixed in master, just use static values in plugin, anyway I think I will improve this after drop webpack@4

evilebottnawi (commented):
And big thanks

PR #718, opened 2020-10-10
chore(deps): schema-utils@3, latest others

AviVahl:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**
- [x] **dependency upgrades**

### Motivation / Use-Case

Upgrade all dependencies to latest stable.
Regenerate lock file from scratch.

### Breaking Changes

None. (node versions already dropped)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,11 +38,11 @@
     "webpack": "^4.0.0 || ^5.0.0"
   },
   "dependencies": {
-    "mem": "^6.1.1",
+    "mem": "^7.0.0-1",
     "memfs": "^3.2.0",
     "mime-types": "^2.1.27",
     "range-parser": "^1.2.1",
-    "schema-utils": "^2.7.1"
+    "schema-utils": "^3.0.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.11.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,11 +38,11 @@
     "webpack": "^4.0.0 || ^5.0.0"
   },
   "dependencies": {
-    "mem": "^6.1.1",
+    "mem": "^7.0.0-1",
```
evilebottnawi, line 41: I think it is not safe to use it right now, it is not stable

AviVahl: it's actually the latest stable. `npm dist-tags mem`:
```
latest: 7.0.0-1
next: 7.0.0-3
```

I can downgrade if you want to wait for a proper semver-ed version.

AviVahl:
Oh, webpack-next fails. I'll debug a bit with the latest rc.

package.json
evilebottnawi: Yep, I think it is mistake

AviVahl:
`webpack-next` is broken due to https://github.com/webpack/webpack/pull/11258
The default value of `publicPath` changed to `"auto"` and webpack-dev-middleware` needs to handle that new possible value, I guess. The stats output also changed, which caused several snapshots to break.

Not sure of the proper way to handle these two.
For the first issue, what does "auto" means for the devMiddleware understanding of urls?
For the second issue, how should snapshots be used when the plugin aims to support two webpack versions with two different outputs.

I'm closing this PR, as I have no clear answer to above questions.

PR #739, opened 2020-10-21
feat: port to typescript, generate .d.ts from src

AviVahl:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

- ports the entire ./src folder to strict typescript.
- the only behavior change  is that `context.stats` is
now `Stats | null` instead of `Stats | null | undefined`.
makes it more consistent with the rest of the context, and it actually starts out as `null`.
- added jsdocs to entire options interface, for better consumer experience.
- removed babel, switched tooling to typescript  (ts-jest, typescript-eslint).
- used a ts namespace to ensure cjs entry has all public types exported.
- used overloading to ensure proper type bahavior of compiler vs multi compiler.
- switch eslintrc to non js file, as IDEs give better intellisense for the rules, etc.
- src is now also published, as .d.ts.map files are generated.
- had to work around several non exported (MultiStats, Watching) webpack types.
- manually verified using with @types/webpack (for webpack@4 users). works, yet easy to break.


AviVahl:
the types and tests are passing, but usage with @types/webpack (webpack@4-typescript consumers) is still not working.
will have to see if I can get it working with both new and older webpack releases.

evilebottnawi (commented):
Sorry, no, only typescript JS docs, like in webpack

package.json
```diff
@@ -45,14 +46,16 @@
...
-    "@babel/cli": "^7.12.1",
-    "@babel/core": "^7.12.3",
-    "@babel/preset-env": "^7.12.1",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
+    "@types/express": "^4.17.8",
+    "@types/jest": "^26.0.15",
+    "@types/mime-types": "^2.1.0",
+    "@types/node": "^10.17.40",
+    "@types/range-parser": "^1.2.3",
```
evilebottnawi, line 55: Most of this types should be in `dependencies`

src/types.ts
```diff
@@ -0,0 +1,127 @@
...
+  watching: MultiWatching | null;
+}
+
+type OutputFileSystem = webpack.Compiler['outputFileSystem'];
+
+// workaround until webpack generates more exact types
+export interface ExtendedOutputFileSystem extends OutputFileSystem {
+  readFileSync(filePath: string): Buffer;
+  statSync(filePath: string): fs.Stats;
+}
```
evilebottnawi, line 108: It should be `OutputFileSystem`

evilebottnawi (commented):
Also a lot of packages, uses our utils, so renaming files is breaking change

src/types.ts
AviVahl: ideally yes. had to invent that to workaround webpack types lacking.

package.json
AviVahl: Installing @types packages for users is a bad practice that tends to break more than it solves. mime-types, range-parser, and jest are only used internally. @types/node shouldn't be installed for user. what if he wants to verify using @types/node@12 or 14...?

AviVahl:
do these packages import from ./src? probably dist, as cjs lives there.
I didn't change the structure of output. all files are named the same in dist.

AviVahl:
> Sorry, no, only typescript JS docs, like in webpack
there's a precedence (webpack-cli) and webpack just added .d.ts to its own package... so the general direction is clear.
there are huge benefits to using native typescript here. javascript (with jsdocs) behavior is not 1:1, and not as strict.

evilebottnawi:
> there's a precedent (webpack-cli) and webpack just added .d.ts to its own package... so the general direction is clear.

webpack-cli team do big mistake and create a lot of problem

> there are huge benefits to using native typescript here. javascript (with jsdocs) behavior is not 1:1, and not as strict.

Sorry, it is our official answer - we prefer JS with typescript JSDocs

I do not want to argue here about the advantages or disadvantages, this is just the official answer

Some packages already ported https://github.com/webpack/schema-utils

AviVahl:
that's a shame. there's a huge community that knows and works with native typescript.
the community for using jsdocs with allowJs/checkJs is much smaller, I feel.
well, I appreciate you took the time to review this. thx.

evilebottnawi:
@AviVahl 

> that's a shame

Stop be toxic?

> there's a huge community that knows and works with native typescript.
> the community for using jsdocs with allowJs/checkJs is much smaller, I feel.

Yes, what the problem with `jsdocs`? We provide types for your exactly the same. Do you understand me correctly?

evilebottnawi:
We just need to used typescript jsdocs comments with the same types and added command https://github.com/webpack/schema-utils/blob/master/package.json#L23, and typescript generates types for our package

AviVahl:
Apologies. Didn't mean to sound toxic.
It's fine, really. There are pros and cons for both approaches.
I am not going to even try to convert this to jsdocs, so re-closing the PR.
Thx again.

evilebottnawi:
> I am not going to even try to convert this to jsdocs, so re-closing the PR.

Why? It would help us all, I see no real problems with this. It won't take many time

AviVahl:
It's extra time to invest in what I consider to be a lesser solution. I rather work on other stuff.

evilebottnawi:
Bad :disappointed: I would be glad to have such help

Issue #738, opened 2020-10-20
ETag support


privatenumber:
- Operating System: Mac
- Node Version: v12.16.2
- NPM Version: v6.14.4
- webpack Version: v4.44.2
- webpack-dev-middleware Version: v3.7.2

### Feature Proposal
Add Etags to assets to prevent re-loading unchanged assets.

### Feature Use Case
Faster page loads by eliminating loads for large files that are unchanged.

This optimization is used in other dev bundlers like [Vite](https://github.com/vitejs/vite) and [Snowpack](https://www.snowpack.dev).

### Other
I implemented this via wrapper [in a separate package](https://github.com/privatenumber/vue-just-ssr/pull/7/files#diff-3979a5e82c57a8459c6e8316962f32833b811d7ca03362e3e3c4319211454a56R105).

Happy to open a PR if the issue is accepted.

evilebottnawi:
@privatenumber Yes, PR welcome

privatenumber:
Is v3 still actively maintained? I didn't see a separate dev branch for it but I'm personally using v3 so I would like to add it there so that I can use it.

If not, I can add it to v4.

evilebottnawi:
Yes, let's add to v4, I will do release in near future (today/tomorrow), also v4 supports webpack@4 and webpack@5

PR #746, opened 2020-10-27
feat: Etag support

privatenumber:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

- Etags: https://github.com/webpack/webpack-dev-middleware/issues/738
- Also, a refactor to use native req/res methods. It was only using express-only functions that were very minor abstractions over the native methods. FYI I'm using [connect](https://www.npmjs.com/package/connect) for my server.

### Breaking Changes

N/A

### Additional Info

Will work on tests tomorrow but opening a PR to get a head start on review and feedback.

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -78,6 +78,7 @@ export default function wrapper(context) {
         const { 'if-none-match': ifNoneMatch } = req.headers;
         if (ifNoneMatch) {
           if (assetEtag === ifNoneMatch) {
+            // eslint-disable-next-line no-param-reassign
             res.statusCode = 304;
             res.end();
             return;
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -56,18 +56,18 @@ export default function wrapper(context) {
         return;
       }
 
-      if (!res.get('Content-Type')) {
+      if (!res.getHeader('Content-Type')) {
         // content-type name(like application/javascript; charset=utf-8) or false
         const contentType = mime.contentType(path.extname(filename));
 
         if (contentType) {
-          res.set('Content-Type', contentType);
+          res.setHeader('Content-Type', contentType);
         }
       }
 
       if (headers) {
         for (const name of Object.keys(headers)) {
-          res.set(name, headers[name]);
+          res.setHeader(name, headers[name]);
         }
       }
 
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -2,11 +2,35 @@ import path from 'path';
 
 import mime from 'mime-types';
 
+import getETag from 'etag';
+
 import getFilenameFromUrl from './utils/getFilenameFromUrl';
 import handleRangeHeaders from './utils/handleRangeHeaders';
 import ready from './utils/ready';
 
+const { hasOwnProperty } = Object.prototype;
+
 export default function wrapper(context) {
+  const { compiler } = context;
+  const etagRegistry = new Map();
+
+  function computeEtags(stats) {
+    const { assets } = stats.compilation;
+    for (const assetId in assets) {
+      if (hasOwnProperty.call(assets, assetId)) {
+        const { existsAt: fsPath } = assets[assetId];
+        const etag = getETag(assets[assetId].source());
+        etagRegistry.set(fsPath, etag);
+      }
+    }
+  }
+
+  if (compiler.hooks.done) {
+    compiler.hooks.done.tap('just-ssr', computeEtags);
+  } else {
+    compiler.plugin('done', computeEtags);
+  }
+
   return async function middleware(req, res, next) {
     const acceptedMethods = context.options.methods || ['GET', 'HEAD'];
     // fixes #282. credit @cexoso. in certain edge situations res.locals is undefined.
@@ -49,6 +73,20 @@ export default function wrapper(context) {
         return;
       }
 
+      const assetEtag = etagRegistry.get(filename);
+      if (assetEtag) {
+        const { 'if-none-match': ifNoneMatch } = req.headers;
+        if (ifNoneMatch) {
+          if (assetEtag === ifNoneMatch) {
+            res.statusCode = 304;
+            res.end();
+            return;
+          }
+        } else {
+          res.setHeader('ETag', assetEtag);
+        }
+      }
+
       try {
         content = context.outputFileSystem.readFileSync(filename);
       } catch (_ignoreError) {
@@ -75,7 +113,7 @@ export default function wrapper(context) {
       content = handleRangeHeaders(context, content, req, res);
 
       // send Buffer
-      res.send(content);
+      res.end(content);
     }
   };
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -2490,7 +2490,11 @@
           "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.13.tgz",
           "integrity": "sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==",
           "dev": true,
-          "optional": true
+          "optional": true,
+          "requires": {
+            "bindings": "^1.5.0",
+            "nan": "^2.12.1"
+          }
         },
         "glob-parent": {
           "version": "3.1.0",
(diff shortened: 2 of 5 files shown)


src/middleware.js
```diff
@@ -2,11 +2,37 @@ import path from 'path';
...
+        }
+      }
+    } catch (_ignoreError) {} // eslint-disable-line no-empty
+  }
+
+  if (compiler.hooks.done) {
+    compiler.hooks.done.tap('webpack-dev-middleware', computeEtags);
+  } else {
+    compiler.plugin('done', computeEtags);
+  }
```
evilebottnawi, line 34: No hooks here

evilebottnawi: We already can load assets here, please look at logic

src/middleware.js
```diff
@@ -2,11 +2,37 @@ import path from 'path';
...
+
 import getFilenameFromUrl from './utils/getFilenameFromUrl';
 import handleRangeHeaders from './utils/handleRangeHeaders';
 import ready from './utils/ready';
 
+const { hasOwnProperty } = Object.prototype;
+
 export default function wrapper(context) {
+  const { compiler } = context;
+  const etagRegistry = new Map();
```
evilebottnawi, line 15: memory leak

evilebottnawi: We need to `WeakMap` and use `assets` source as key

src/middleware.js
```diff
@@ -49,6 +75,19 @@ export default function wrapper(context) {
         return;
       }
 
+      const assetEtag = etagRegistry.get(filename);
+      if (assetEtag) {
```
evilebottnawi, line 79: We need option to enable/disable it, in future we can enable it by defualt

src/middleware.js
```diff
@@ -2,11 +2,37 @@ import path from 'path';
...
+  const { compiler } = context;
+  const etagRegistry = new Map();
+
+  function computeEtags(stats) {
+    try {
+      const { assets } = stats.compilation;
+      for (const assetId in assets) {
+        if (hasOwnProperty.call(assets, assetId)) {
+          const { existsAt: fsPath } = assets[assetId];
+          const etag = getETag(assets[assetId].source());
```
evilebottnawi, line 23: `assets[assetId]` is key for `WeakMap`

evilebottnawi (commented):
Also we have `immutable` flag in `assetInfo`, this flag assets will do be changed never

src/middleware.js
privatenumber: AFAIK, Weakmaps don't allow primitives (eg. strings) to be used as keys

src/middleware.js
privatenumber: Hmm, I'll need to be able to "fetch" this etag value whenever a new request comes in, and if we use the Asset instance as a key, I'm not sure if I'll be able to query the etag.

Would clearing the Map work?

src/middleware.js
evilebottnawi: Why you need `iterate`?

src/middleware.js
evilebottnawi: No, it is still memory leak, you should not use primitives for weak amp

src/middleware.js
evilebottnawi: Sorry, no, clearing map is not fix memory leaking

src/middleware.js
evilebottnawi: Anyway if you need this, you have `context`, put this logic inside `done` hooks and collect them `context.etags`

privatenumber: Sounds good.

I want to iterate a list of distribution files to efficiently calculate etags. If I only have the memfs instance, I'd have to traverse the entire fs, which isn't optimal.

src/middleware.js
privatenumber: Resolved the other thread about this so all conversation is in this thread.

Can you explain why there would still be a memory leak if unused etags are being cleaned up per build?

evilebottnawi: > Can you explain why there would still be a memory leak if unused etags are being cleaned up per build?

Because we create `etagRegistry` on middleware start so next compilation will put new files again and again, and if you use `name.[contenthash][ext]` you will increase size infinitely, it will be difficult to debug, but this is a potential leak, using `WeakMap` help to solve it, when asset was removed from `comilation.assets` it will be removed from `WeakMap`, so we free memory

privatenumber: I see, thanks for explaining.

That's addressed by [`etagRegistry.clear()`](https://github.com/webpack/webpack-dev-middleware/pull/746/files#diff-9a117d52a8a2cb3e5f498575bdfdea4dc32418180c09029454fbaf7f95ebe5a1R19) on every build completion because `.clear()` removes all entries in a Map.


I'm wondering how you imagine using a WeakMap is possible? I would jump to using it if it were possible to get the asset object (they key) when a new HTTP request comes in [here](https://github.com/webpack/webpack-dev-middleware/pull/746/files#diff-9a117d52a8a2cb3e5f498575bdfdea4dc32418180c09029454fbaf7f95ebe5a1R77).

privatenumber:
As I work on the tests, I'm just realizing that [express adds etags by default](https://github.com/expressjs/express/blob/master/lib/response.js#L201).

Feels redundant to make it disabled by default on the middleware, and this also makes it difficult to test because etags are on even if the middleware option is disabled.

Do you mind if I make this a PR against https://github.com/webpack/webpack-dev-middleware/pull/747? That way I can pull in the vanilla HTTP version and use connect in the test.

src/middleware.js
evilebottnawi: > That's addressed by etagRegistry.clear() on every build completion because .clear() removes all entries in a Map.

it was reduce performance, we should avoid it

> I'm wondering how you imagine using a WeakMap is possible? I would jump to using it if it were possible to get the asset object (they key) when a new HTTP request comes in here.

store source of assets and use it as key

evilebottnawi:
> As I work on the tests, I'm just realizing that express adds etags by default.

Yes, we will enable it by default, but we really need a option to disable this


> Do you mind if I make this a PR against #747? That way I can pull in the vanilla HTTP version and use connect in the test.

Yes

evilebottnawi:
I studied all the material and other implementation, so I want to implement options:

- cacheControl
- etag
- lastModified

It should be not hard, WIP on this, but I think release will be on the next week only

evilebottnawi:
Anyway thanks for good idea

privatenumber:
@evilebottnawi 

I thought about time-based caching (cache-control/if-modified-since) as well but concluded it wouldn't be as useful as a content-based one (etag) in a rapid development environment.

I imagine it would actually lead to:
- unexpected behavior (changed code not getting immediately fetched)
- users not understanding the feature and disabling caching in the browser
- people turning it off/not using it

evilebottnawi:
I will provide some docs on this to help developers, anyway etag is very easy to implement, so it will be implemented in near future and enabled by default

src/middleware.js
```diff
@@ -2,11 +2,35 @@ import path from 'path';
...
+    etagRegistry = new Map();
+    context.compiler.hooks.done.tap('webpack-dev-middleware', (stats) => {
+      etagRegistry.clear();
+
+      try {
+        const { assets } = stats.compilation;
+        for (const assetId in assets) {
+          if (hasOwnProperty.call(assets, assetId)) {
+            const { existsAt: fsPath } = assets[assetId];
+            const etag = getETag(assets[assetId].source());
```
privatenumber, line 26: Seems this is no longer possible in Webpack 5, due to error:

> <e> [webpack-dev-middleware] Error: Content and Map of this Source is not available (only size() is supported)
> <e>     at SizeOnlySource._error (/node_modules/webpack-sources/lib/SizeOnlySource.js:16:10)
> <e>     at SizeOnlySource.source (/node_modules/webpack-sources/lib/SizeOnlySource.js:26:14)

@alexander-akait Do you have a recommended work around for this?

src/middleware.js
```diff
@@ -2,11 +2,35 @@ import path from 'path';
...
+  if (context.options.etags) {
+    etagRegistry = new Map();
+    context.compiler.hooks.done.tap('webpack-dev-middleware', (stats) => {
+      etagRegistry.clear();
+
+      try {
+        const { assets } = stats.compilation;
+        for (const assetId in assets) {
+          if (hasOwnProperty.call(assets, assetId)) {
+            const { existsAt: fsPath } = assets[assetId];
```
privatenumber, line 25: Seems `existsAt` doesn't exist in Webpack 5 either

src/middleware.js
privatenumber: I ended up leveraging `chunkHashs`, which is actually more convenient because I can just use the hash as the etag instead of generating one from the file source:

```js
const { chunkHashs: assetHashes } = stats.compilation.records;
for (const assetId in assetHashes) {
	if (hasOwnProp(assetHashes, assetId)) {
		etagRegistry.set(assetId, JSON.stringify(assetHashes[assetId]));
	}
}
```

If there's an alternative you recommend, LMK.

alexander-akait: /cc @sokra What do you think?

PR #747, opened 2020-10-27
refactor: use native req/res methods

privatenumber:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
The code was using express specific methods for very simple things (eg. getting/setting header).

I use [connect](https://www.npmjs.com/package/connect) to keep my server light-weight, and others may be using frameworks like koa.

This change refactors the express method calls to use native Node.js HTTP request/response methods instead.


evilebottnawi (changes requested):
We need to setup tests for connect in this case, we can't just change code because something use it, it can breaking other compatibles

evilebottnawi:
I can help you with it, should be not hard

ylemkimon:
Note this is a revert of #726.

evilebottnawi:
Yes, so I ask to add tests and we should create `compatibility.js` file and store these hacks there

privatenumber:
I can wrap the entire `middleware.test.js` with an iteration so they all run for `express` and `connect`, but this might add a lot of time to tests because it's a very large test file.

What do you think of just swapping `express` with `connect` in the test environment?

Since `connect` is just a middleware framework and doesn't add custom request/response methods like `express` does, using the underlying native HTTP request/response methods and testing with `connect` will also mean it works for `express`. (middlewares designed for `connect` can always be used by `express`, eg. [`compression`](https://www.npmjs.com/package/compression)).

evilebottnawi:
> What do you think of just swapping express with connect in the test environment?

Bad idea, we should we should test on what it works on

> Since connect is just a middleware framework and doesn't add custom request/response methods like express does, using the underlying native HTTP request/response methods and testing with connect will also mean it works for express. (middlewares designed for connect can always be used by express, eg. compression).

yes, but we need to add it for all tests, otherwise we can't guarantee compatibility

privatenumber:
How would you like to add tests for both? (eg. should I wrap the entire middleware tests so they run for both express and connect?)

Also, supporting connect is effectively supporting a vanilla Node.js HTTP server as well, if you're interested in adding tests for that as well.

evilebottnawi:
> How would you like to add tests for both? (eg. should I wrap the entire middleware tests so they run for both express and connect?)

Ideally - yes

> Also, supporting connect is effectively supporting a vanilla Node.js HTTP server as well, if you're interested in adding tests for that as well.

Yes :+1:

privatenumber:
@evilebottnawi 

I added an outer iteration in the middleware test so that they're executed for `connect` and `express`. There's a few tests that fail though, but only when I run them both. If I comment out either `connect` or `express` they all succeed.

I couldn't find out what was causing this. Do you mind looking into this?

Also looked into adding a vanilla HTTP server, but I may have over-spoke. Adding middleware support wasn't as easy as I imagined... the whole [`connect` module is doing that](https://github.com/senchalabs/connect/blob/master/index.js) after all, so probably not necessary anyway.

ylemkimon:
> BTW I still think it's kinda pointless to test for express if we're testing with connect, and we're not using any express features. At most, I think just add a smaller express smoke test. Maybe you'll agree as you checkout the code.

[webpack-dev-server](https://github.com/webpack/webpack-dev-server) uses express for the framework.

privatenumber:
Thanks for chiming in @ylemkimon

This PR removes dependency on express-only API so that it can be HTTP framework agnostic. Instead of using the express API, this middleware now only uses [native Node.js HTTP API](https://nodejs.org/api/http.html). Therefore, this middleware now works with _any_ HTTP framework that expose native request/response instances because we're no longer interacting with any framework code. The only dependency now is the "HTTP middleware paradigm".

You can confirm this by swapping `express` with `connect` in the tests. They will all pass because the code is now framework agnostic.

`connect` is just the "middleware" feature of `express`, and because of its minimal approach, it makes it easier to test and keep a stricter scope. For example, express automatically adds etags, which interferes with testing the etag feature that I'm working on [here](https://github.com/webpack/webpack-dev-middleware/pull/746).


So, testing with `express` in addition to `connect` is almost like, making a module that interacts with the DOM, and testing it with Preact and React via mounted hook... as long as it's interacting with the DOM, it's going to work so you don't need to test it with every JS framework out there.

src/middleware.js
```diff
@@ -56,26 +56,31 @@ export default function wrapper(context) {
...
         // content-type name(like application/javascript; charset=utf-8) or false
         const contentType = mime.contentType(path.extname(filename));
-
-        if (contentType) {
-          res.set('Content-Type', contentType);
-        }
+        res.setHeader(
+          'Content-Type',
+          contentType || 'application/octet-stream'
+        );
```
evilebottnawi, line 65: We should **always** use official middleware API, so for example we shoudl use `express` API, for connect (no example API) we should use http module API, why? Because other middlewares/tools can add additional logic for built-in API, and we will not working properly with them

evilebottnawi: So we need to create `compatibility` file and put some utils there, like:
```

clarkdo:
Thanks for this pr, look forward to the release.

src/middleware.js
privatenumber: I agree with your explanation, and that's why this middleware can't be used by Koa, even though they expose native request/response.

However, if we scope our support level to vanilla HTTP, express, connect, and other frameworks that don't add extra logic, then we don't need to do that. We can't possibly support all frameworks anyway.

evilebottnawi: So let's take my suggestion, it is easy and will not confusing for future refactoring

privatenumber: If the support level is strictly express and connect, I think it's unnecessary.

Going back to the case with [`compression`](https://www.npmjs.com/package/compression), I'm not convinced why a third-party middleware requires this abstraction when a first-party middleware doesn't.

privatenumber:
@evilebottnawi 

Following up with https://github.com/webpack/webpack-dev-middleware/pull/747#issuecomment-718387867

I couldn't find out what was causing this. Do you mind looking into this?

src/middleware.js
evilebottnawi: Can you clarify?

evilebottnawi:
Okay, I am start to working on it, want to separate it on two commits - prepare tests + compatibility fixes to avoid dirty git history

evilebottnawi:
Fixed https://github.com/webpack/webpack-dev-middleware/pull/760

@privatenumber big thanks for helping and investigating, I will do release tomorrow, the next target - `ETag` header and maybe other cache headers

privatenumber:
Thanks for getting it in @evilebottnawi. Looking forward to being able to use it!

Not a big deal because I primarily did this for the feature, but for future reference, I would appreciate it if my commits could've somehow been integrated into your PR and the repository. I put in a lot of hours into this but I won't be credited as a [contributor](https://github.com/webpack/webpack-dev-middleware/graphs/contributors).

PR #804, opened 2020-12-21
fix(output-file-system): change from backslash to forward slash in join

soilSpoon:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

In `webpack`, if there is a `join` method of `outputFileSystem`, use the `join` method. However, if you use the `join` method in `windows os`, a problem occurs when the path is relative.
Ex)
before `join`:
`/app/js`
after `join`:
`\app\join`

If it changes as above, the following error occurs.

```
D:\foo\node_modules\webpack\lib\util\fs.js:141
                throw new Error(
                ^

Error: \js\app.js is neither a posix nor a windows path, and there is no 'dirname' method defined in the file system
    at dirname (D:\foo\node_modules\webpack\lib\util\fs.js:141:9)
    at D:\foo\node_modules\webpack\lib\Compiler.js:790:19
    at arrayIterator (D:\foo\node_modules\neo-async\async.js:3467:9)
    at timesSync (D:\foo\node_modules\neo-async\async.js:2297:7)
    at Object.eachLimit (D:\foo\node_modules\neo-async\async.js:3463:5)
    at emitFiles (D:\foo\node_modules\webpack\lib\Compiler.js:536:13)
    at D:\foo\node_modules\webpack\lib\util\fs.js:182:5
    at Immediate.<anonymous> (D:\foo\node_modules\memfs\lib\volume.js:684:17)
    at processImmediate (internal/timers.js:461:21)
```

This PR solves that problem.

### Breaking Changes

The `join` method of `outputFileSystem` returns all backslashes replaced with slashes.

### Additional Info
https://github.com/webpack/webpack/blob/master/lib/util/fs.js#L133-L145

### DIFF
--- a/src/utils/setupOutputFileSystem.js
+++ b/src/utils/setupOutputFileSystem.js
@@ -27,7 +27,9 @@ export default function setupOutputFileSystem(context) {
   } else {
     outputFileSystem = createFsFromVolume(new Volume());
     // TODO: remove when we drop webpack@4 support
-    outputFileSystem.join = path.join.bind(path);
+    const newPath = Object.assign({}, path);
+    newPath.join = (...files) => path.join(...files).replace(/[\\/]+/g, '/');
+    outputFileSystem.join = newPath.join.bind(newPath);
   }
 
   const compilers = context.compiler.compilers || [context.compiler];


alexander-akait:
Please provide problem

alexander-akait:
Also this code for webpack@4

src/utils/setupOutputFileSystem.js
```diff
@@ -27,7 +27,9 @@ export default function setupOutputFileSystem(context) {
   } else {
     outputFileSystem = createFsFromVolume(new Volume());
     // TODO: remove when we drop webpack@4 support
-    outputFileSystem.join = path.join.bind(path);
+    const newPath = Object.assign({}, path);
+    newPath.join = (...files) => path.join(...files).replace(/[\\/]+/g, '/');
```
alexander-akait, line 31: Using regexp to change `\` on `/` is bad idea

soilSpoon:
> Please provide a problem

Can you please let me know what more explanations are needed?

src/utils/setupOutputFileSystem.js
soilSpoon:
```suggestion
    newPath.join = (...files) => path.join(...files).replaceAll('\\', '/');
```
Would this be better?

alexander-akait:
Please show me a problem why you need to change it

alexander-akait:
Simple reproducible test repo

soilSpoon:
@alexander-akait Sorry i think i was mistaken

PR #809, opened 2021-01-01
fix: skip content-type header on unknown types

### LINKED ISSUE #354: bug: unknown mime type sets Content-Type as 'null; charset=UTF-8'
* Operating System: Mac OS X
* Node Version: v8.11.3
* NPM Version: 6.5.0
* webpack version: 4.18.0

- [X] This is a **bug**

### Code

```js
return {
    entry: entries,
    devtool: isProduction ? 'source-map' : 'nosources-source-map',
    mode: process.env.NODE_ENV || 'development',
    devServer: {
      host: '0.0.0.0',
      hot: false,
      inline: false,
    },
    output: {
      filename: '[name].min.js',
      path: path.resolve(__dirname, DEST, folder),
    },
    module: {
      rules: [{
        test: /\.js$/,
        include: [
          path.resolve(__dirname, 's3', folder),
        ],
        use: {
          loader: 'babel-loader',
          options: {
            presets: [
              ['@babel/preset-env'],
            ],
          },
        },
      }, {
        test: /\.scss$/,
        use: [
          'file-loader?name=[name].min.css',
          'extract-loader',
          'css-loader?import=false',
          {
            loader: 'postcss-loader',
            options: {
              plugins: () => [
                cssnano(),
              ],
            },
          },
          'sass-loader',
        ],
      }, {
        test: /_.+\.pug$/,
        use: [
          'html-loader',
          'pug-html-loader',
        ],
      }, {
        test: /index\.pug$/,
        use: [
          'file-loader?name=index.html',
          'extract-loader',
          'html-loader',
          'pug-html-loader',
        ],
      },
      {
        test: /\.pug$/,
        exclude: /(_.+)|(index)\.pug$/,
        use: [
          'file-loader?name=[name]',
          'extract-loader',
          'html-loader',
          'pug-html-loader',
        ],
      }],
    },
````

### Expected Behavior

Unknown mime-types should return a blank media-type in Content-Type

### Actual Behavior

`media-type` is being listed as null

Create a webpack package that outputs any file without an extension.

----------

The issue is in `middleware.js`

https://github.com/webpack/webpack-dev-middleware/blob/e6cfe7c10c423964a8b3373204a1cec27c1a4c91/lib/middleware.js#L72

`mime.getType(filename);` will return null if it cannot lookup the media-type.

https://github.com/broofa/node-mime/blob/a1d884c1672f34031af15718b25b93dee00dc3ad/Mime.js#L82

When `'; charset=UTF-8'` is added to null, the result is `Content-Type: null; charset=UTF-8`.

https://github.com/webpack/webpack-dev-middleware/blob/e6cfe7c10c423964a8b3373204a1cec27c1a4c91/lib/middleware.js#L76

Different browsers treat `null` differently. Chrome, Safari, and Firefox will try to interpret the file (which in my use case is `text/html`). IE and Edge will not and treat it as `application/octet-stream` which will cause a download prompt to appear.  Using `*/*` will produce the same result.

If we don't use 'null', and instead, leave the media-type blank, all the above browsers will try to interpret the file when `Content-Type: ; charset=UTF-8` is used. Since I'm using it for HTML file, it gets interpreted correct by IE and Edge. That's the expected behavior according to RFC 7231:

>If a Content-Type header field is not present, the recipient MAY either assume a media type of "application/octet-stream" or examine the data to determine its type.

https://tools.ietf.org/html/rfc7231#section-3.1.1.5

clshortfuse:
Fixes #354

* Only set 'Content-Type' header if mime type is known.
* Remove testing for 'Content-Type' of unknown types.
* Do not test against `application/octet-stream'.

Unknown media types should have content-type blank. Because
frameworks handle Content-Types differently, it is not possible to
standardize testing at the moment. Tests against
`application/octet-stream` can create false positives with Express
because of this.

https://tools.ietf.org/html/rfc7231#section-3.1.1.5
https://github.com/broofa/mime/issues/139

This PR contains a:

- [x] **bugfix**
- [x] **test update** 

### Breaking Changes

Unknown mime types (no extension or unknown extension) no longer produce a `Content-Type` with Connect/NodeJS API.

### Additional Info

https://tools.ietf.org/html/rfc7231#section-3.1.1.5
https://github.com/broofa/mime/issues/139

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -65,16 +65,17 @@ export default function wrapper(context) {
         // content-type name(like application/javascript; charset=utf-8) or false
         const contentType = mime.contentType(path.extname(filename));
 
-        // Express API
-        if (res.set && contentType) {
-          res.set('Content-Type', contentType);
-        }
-        // Node.js API
-        else {
-          res.setHeader(
-            'Content-Type',
-            contentType || 'application/octet-stream'
-          );
+        // Only set content-type header if media type is known
+        // https://tools.ietf.org/html/rfc7231#section-3.1.1.5
+        if (contentType) {
+          // Express API
+          if (res.set) {
+            res.set('Content-Type', contentType);
+          }
+          // Node.js API
+          else {
+            res.setHeader('Content-Type', contentType);
+          }
         }
       }
 
--- a/test/middleware.test.js
+++ b/test/middleware.test.js
@@ -290,7 +290,6 @@ describe.each([
           request(app)
             .get('/unknown')
             .expect('Content-Length', fileData.byteLength.toString())
-            .expect('Content-Type', 'application/octet-stream')
             .expect(200, done);
         });
       });
@@ -584,7 +583,6 @@ describe.each([
             urls: [
               {
                 value: 'filename-name-with-dots/mono-v6.x.x',
-                contentType: 'application/octet-stream',
                 code: 200,
               },
             ],
@@ -595,7 +593,6 @@ describe.each([
             urls: [
               {
                 value: 'noextension',
-                contentType: 'application/octet-stream',
                 code: 200,
               },
             ],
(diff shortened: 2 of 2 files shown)


alexander-akait (approved):

alexander-akait:
Big thanks for the work

PR #817, opened 2021-01-13
fix: use webpack logger to print stats

clarkdo:
This PR contains a:

- [x] **bugfix**

### Motivation / Use-Case

The latest `webpack-dev-middleware` changed from logger to console.log for printing stats, it breaks usage that someone is using customized webpack logger by `infrastructureLog` hook or `infrastructureLogging` config.

I've tested locally with logger.log, it works fine with coloured stats log, I'm not sure if using console.log is intentional or not, if there is any valid reason, please feel free to let me know.

### DIFF
--- a/src/utils/setupHooks.js
+++ b/src/utils/setupHooks.js
@@ -88,8 +88,7 @@ export default function setupHooks(context) {
 
       // Avoid extra empty line when `stats: 'none'`
       if (printedStats) {
-        // eslint-disable-next-line no-console
-        console.log(printedStats);
+        logger.log(printedStats);
       }
 
       // eslint-disable-next-line no-param-reassign


alexander-akait (changes requested):
No, logger output stats to stderr, it is wrong, it should be `stdout`, and stats !== logger

clarkdo:
@alexander-akait So is there anyway I can disable printing stats in webpack-dev-middleware but not disable webpack stats(I need stats info for other plugins) ?

alexander-akait:
```{ stats: isDevServer ? 'none' : 'normal' }```, if you use `webpack-cli`, you can do this:
```
module.exports = (env) => { 
  stats: env.WEBPACK_SERVER ? 'none' : 'normal'
};
```

clarkdo:
Sorry, I mean I'm using other plugins for logging build info (include prettified stats), so I want to disable the duplicate stats printed in `webpack-dev-middleware`

alexander-akait:
How you use `webpack-dev-middleware`?

clarkdo:
Nuxt.js is using https://github.com/nuxt-contrib/webpackbar for webpack progress and stats logging, so we want to ignore stats printing from `webpack-dev-middleware`.

alexander-akait:
If you use custom stats output, just use `stats: none` or rewrite value in the plugin

alexander-akait:
@clarkdo What is solution you choose? In theory we can return the `stats` option to webpack-dev-middleware, but I need more info

clarkdo:
I’m trying to switch off stats config in webpack config and let webpackbar print the stats info, but it would be great if this middleware can have similar config as [devServer.stats](https://webpack.js.org/configuration/dev-server/#devserverstats-)

alexander-akait:
@clarkdo Anyway, how `webpack-dev-middleware` run in nuxt? Do you have ability to set the `stats` option for `webpack-dev-middleware`?

clarkdo:
Sorry, here is how we run it https://github.com/nuxt/nuxt.js/blob/dec8f99fc3ffbdcdf731b671b40ef3fd199e7b6b/packages/webpack/src/builder.js#L181

alexander-akait:
@clarkdo Great, please open the new issue and I will do it tomorrow, thanks

clarkdo:
Issue created https://github.com/webpack/webpack-dev-middleware/issues/818

PR #876, opened 2021-04-15
Don't read full file if range header is present

JonnyBurger:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I am reading in a 2GB file for [Remotion](https://remotion.dev). This is a framework for making videos in React, and the end product is not shipped to a website, but written to an MP4 file. Therefore it's not a problem that we import these large files in Webpack.

However, when the browser is loading a video, and we seek forward, then a range header is sent to Webpack Dev Server. Unfortunately it is slow, because WDS is still loading the full file. I adapted the logic from `serve-handler` instead so instead of reading the full 2GB file synchronously, it uses a Read Stream.

### Breaking Changes

Aiming to have no breaking changes.

### Additional Info

Will try to finish this PR when I get time. Might take a while!


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "webpack-dev-middleware",
+  "name": "@jonny/webpack-dev-middleware",
```
alexander-akait, line 2: Please fix it

alexander-akait (changes requested):
Also please add tests

package.json
JonnyBurger: Will clean it up and mark the PR as ready!

JonnyBurger:
@alexander-akait 

I finally got around to making this PR, all the tests are passing now!
There are already tests for the range header, I added some more assertions that the actual response has the correct length and content.

JonnyBurger:
@alexander-akait I think code coverage decrease should be fixed. Can you approve the CI runs? If everything goes green, it's ready for review.

PR #883, opened 2021-04-24
refactor: log publicPath info

snitin315:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

For https://github.com/webpack/webpack-dev-server/pull/3218

### Breaking Changes
Maybe

### Additional Info
None

### DIFF
--- a/src/utils/getPaths.js
+++ b/src/utils/getPaths.js
@@ -12,6 +12,10 @@ export default function getPaths(context) {
       ? compilation.getPath(compilation.outputOptions.publicPath)
       : '';
 
+    if (publicPath) {
+      context.logger.info(`webpack output is served from "${publicPath === 'auto' ? '/' : publicPath}" URL`);
+    }
+
     publicPaths.push({ outputPath, publicPath });
   }
 


src/utils/getPaths.js
```diff
@@ -12,6 +12,10 @@ export default function getPaths(context) {
       ? compilation.getPath(compilation.outputOptions.publicPath)
       : '';
 
+    if (publicPath) {
+      context.logger.info(`webpack output is served from "${publicPath === 'auto' ? '/' : publicPath}" URL`);
+    }
```
alexander-akait, line 17: It will be spam on each request, we need move it only on constructor place

snitin315: I will update 👍🏻

alexander-akait:
I was wrong, I think we don't need this logging, it is weird, because `publicPath` can be `Function` or/and can contain `[hash]`/`[fullhash]`, so our logging will be weird or invalid, I think docs are enough

snitin315:
So we just remove this logging from dev-server ?

alexander-akait:
Yes, I think most of developers see these messages like spam, so let's remove it, I don't think we create problem here removing this, also plugins can change publicPath too, so we know real publicPath only after compilation

snitin315:
Makes sense 👍

Issue #749, opened 2020-10-27
Add support to HEAD requests and Last-Modified response header


watery:
- Operating System: Docker with `node:10.15.3-alpine` image
- Node Version: 10.15.3
- NPM Version:
- webpack Version: from react-scripts @ 2.1.8
- webpack-dev-middleware Version: from react-scripts @ 2.1.8

### Feature Proposal
Make the dev-server reply to HEAD requests on resources where it already replies to GET requests; and add the `Last-Modified` response header if it's not already supported.

### Feature Use Case
I'm doing some experiments in a custom project that rely on `HEAD` requests and the `Last-Modified` response header to detect a (React) app updated deploy. I'm already obtaining some results in a production build served by a full-featured webserver (nginx), and it would be nice to have here aswell, expecially in cases where the browser or page is closed and reopened between source code updates - which if I'm not mistaken, detaches the client code from the dev-server making the former lose the *recompiled* event that triggers a complete page reload.

The latter is the case with `create-react-app` applications, that do not provide a debug build for deployment to staging servers  (see [Support for staging builds #790](https://github.com/facebook/create-react-app/issues/790)); thus running a dev-server instance on a staging server is the only option to let testers / non-developers to access the application for review.

I actually noticed that a `HEAD` request returns 404 whereas a GET on the same resource returns 200, so maybe this is only a matter of supporting `HEAD` requests too.

I've been directed here from [this issue of mine](https://github.com/webpack/webpack-dev-server/issues/2797) from the dev-server subproject. I think I copied over all the relevant information, but I'm linking it just in case.

evilebottnawi:
> I'm already obtaining some results in a production build served by a full-featured webserver (nginx), and it would be nice to have here aswell, expecially in cases where the browser or page is closed and reopened between source code updates - which if I'm not mistaken, detaches the client code from the dev-server making the former lose the recompiled event that triggers a complete page reload.

Last-Modified is hot help here, webpack-dev-server always reload the page when new compilation happens, I think you do not quite understand how it works. What is the real use case for this feature?

evilebottnawi:
Please provide more information

evilebottnawi:
And we already have `HEAD` support

watery:
I need some time to elaborate more on my inital explanation; but I'd like to know more about this:

> And we already have `HEAD` support

Good!
[...]

PR #890, opened 2021-04-28
feat: headers support function 'Last-Modified': () => new Date() (#749)

liuxian1990:
#749 I think we can solve this problem in this way

```js
webpackDevMiddleware(compiler, {
  headers: {
    'Last-Modified': () => new Date()
  }
})
```

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -83,13 +83,18 @@ export default function wrapper(context) {
         const names = Object.keys(headers);
 
         for (const name of names) {
+          let value = headers[name]
+
+          if (typeof value === 'function') {
+            value = value()
+          }
           // Express API
           if (res.set) {
-            res.set(name, headers[name]);
+            res.set(name, value);
           }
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
I think better to allow `headers` will be `function`

liuxian1990:
Good idea

Can it fit into the schedule

alexander-akait:
Yes, feel free to improve, even more, maybe we can send `req`/`res`/etc as arguments, so you can create more complex things

liuxian1990:
Agree with.

I really need this feature and look forward to the update

liuxian1990:
I don't know much English, and this is my first time to submit PR.

Do you mean I will continue to improve, or you will improve yourself

alexander-akait:
> Do you mean I will continue to improve, or you will improve yourself

You can continue improve, it is easy to implement

liuxian1990:
ok

PR #968, opened 2021-07-12
replace mem with better, smaller alternative

43081j:
This PR contains a:

- [x] **metadata update**

I suppose its "metadata", manifest updates.

### Motivation / Use-Case

mem pulls in 3-4 deep of dependencies for something we can achieve with a single file of code.

this just drops it for a more sensible, dependency-free package that does the same job and does it faster too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "license": "MIT",
       "dependencies": {
         "colorette": "^1.2.2",
-        "mem": "^8.1.1",
+        "fast-memoize": "^2.5.2",
         "memfs": "^3.2.2",
         "mime-types": "^2.1.31",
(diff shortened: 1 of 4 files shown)


alexander-akait (commented):
I think we can just drop this package in favor built-in function, it is not hard to implement

43081j:
works for me.

we only use it here so a map of string -> url would do i think.

mind taking a look at my last commit?

tbh we probably don't even need to cache it anymore, creating an instance of the newer `URL` class isn't likely to be slow

43081j:
have rebased onto master and simply dropped the dependency, using a weakmap instead.

let me know if it makes sense / any changes

PR #1002, opened 2021-08-11
feat: don't read full file if range header is present

JonnyBurger:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [x] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I am reading in a 2GB file for [Remotion](https://remotion.dev). This is a framework for making videos in React, and the end product is not shipped to a website, but written to an MP4 file. Therefore it's not a problem that we import these large files in Webpack.

However, when the browser is loading a video, and we seek forward, then a range header is sent to Webpack Dev Server. Unfortunately it is slow, because WDS is still loading the full file. I adapted the logic from `serve-handler` instead so instead of reading the full 2GB file synchronously, it uses a Read Stream.

### Breaking Changes

No breaking changes.

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -48,7 +48,7 @@ export default function wrapper(context) {
         headers = headers(req, res, context);
       }
 
-      let content;
+      let fileSize;
 
       if (!filename) {
         await goNext();
@@ -56,7 +56,7 @@ export default function wrapper(context) {
       }
 
       try {
-        content = context.outputFileSystem.readFileSync(filename);
+        fileSize = context.outputFileSystem.lstatSync(filename).size;
       } catch (_ignoreError) {
         await goNext();
         return;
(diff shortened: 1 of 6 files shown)


alexander-akait (commented):
What we need to finish?

JonnyBurger:
@alexander-akait Sorry I forgot to remove the WIP notice. The PR is ready to be reviewed with no further changes intended on my side.

alexander-akait:
> I adapted the logic from serve-handler instead so instead of reading the full 2GB file synchronously, it uses a Read Stream.

Can you provide link on the original code?

JonnyBurger:
> Can you provide link on the original code?

It's here: 
https://github.com/vercel/serve-handler/blob/c7a40435bc28b420a725e6dc1e9565d5b845dee2/src/index.js#L707-L733

JonnyBurger:
Is there anything left I can do to help this PR progress?

alexander-akait:
@JonnyBurger Sorry for delay, still in my TODO, hope I will merge it on this week

PR #1195, opened 2022-02-06
fix: types

snitin315:
This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Fix https://github.com/webpack/webpack-dev-middleware/issues/1194

### Breaking Changes

None

### Additional Info

No

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@types/connect": "^3.4.35",
     "@types/express": "^4.17.13",
     "@types/mime-types": "^2.1.1",
+    "@types/node": "^12.20.43",
     "@webpack-contrib/eslint-config-webpack": "^3.0.0",
     "babel-jest": "^27.4.6",
     "chokidar": "^3.5.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@types/connect": "^3.4.35",
         "@types/express": "^4.17.13",
         "@types/mime-types": "^2.1.1",
+        "@types/node": "^12.20.43",
         "@webpack-contrib/eslint-config-webpack": "^3.0.0",
         "babel-jest": "^27.4.6",
         "chokidar": "^3.5.1",
@@ -3308,9 +3309,9 @@
       "dev": true
     },
     "node_modules/@types/node": {
-      "version": "17.0.14",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-17.0.14.tgz",
-      "integrity": "sha512-SbjLmERksKOGzWzPNuW7fJM7fk3YXVTFiZWB/Hs99gwhk+/dnrQRPBQjPW9aO+fi1tAffi9PrwFvsmOKmDTyng==",
+      "version": "12.20.51",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.20.51.tgz",
+      "integrity": "sha512-anVDMfReTatfH8GVmHmaTZOL0jeTLNZ9wK9SSrQS3tMmn4vUc+9fVWlUzAieuQefWDyWUz4Z3aqXxDgO1VsYjg==",
       "dev": true
     },
     "node_modules/@types/normalize-package-data": {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "webpack": "^4.0.0 || ^5.0.0"
   },
   "dependencies": {
+    "@types/node": "^12.20.43",
```
alexander-akait, line 50: I think we should avoid it, because it can be problem in future, ideally I think we can just copy `statSync` declaration in our types with todo comment

gbiryukov: 1. probably it will not be that simple because dev-middleware implicitly imports from `@types/node` several other very complex types like `import("http").IncomingMessage`. Other examples: `import("http").ServerResponse` or `typeof import("fs").readFileSync`. 
2. seems dependency should go to `devDependencies` and should allow newer major versions as well

would be great to move forward this PR because it blocks TS users from updating

alexander-akait: Why not update `@types/node` locally?

gbiryukov: it will be breaking change and will require minimal supported `node` version to be `16` for `webpack-dev-middleware`.
you can't safely use `node@16` typings with `node@12` codebase because it includes interface and feature changes from v16.

primary issue remains: `webpack-dev-middleware` does not list dependency it actually uses: `@types/node@X.X.X`

TheLarkInn: Let's revisit this review @alexander-akait. What are we proposing as the better alternative solution here. If types are autogenerated it doesn't make sense to patch the `statSync` declaration.

iclanton: This should be a `peerDependency` and a `devDependency`, it should be an optional peer dependency, and the version should be `"*"` to avoid side-by-side versions in repos with a different major Node version (or older minor/patch version).

iclanton: Looks like this problem is solved if you just make this a `devDependency` of this project.

alexander-akait: To be honestly there is simple solution - just inline `import("fs").StatSyncFn` in our code (i.e. copy paste types) and add todo - `// TODO return import("fs").StatSyncFn when we update Node.js`, so we will not have problems like this, I am not a fun to put `@types/node` in deps - more package size, also we don't have magic here, just use Node.js API when it is avaliable, not sure we should increase node_modules size and reduce install time, also putting it in types can have side effects for projects without installed `@types/node`

iclanton: @alexander-akait - I'm suggesting putting it in the **`devDependencies`** of this project, so an older version of `@types/node` is available at build time.

iclanton: Running `npm install` on the current `master` branch gives a `node_modules` folder with a hoisted `@types/node@17.0.14` dependency because some other dependency is pulling that in. Because of this hoisted dependency, the TypeScript compiler is able to get at the Node types, and the version it's getting is `17.0.14`. 

The `@typedef` of the `OutputFileSystem` ([here](https://github.com/webpack/webpack-dev-middleware/blob/master/src/index.js#L48)) looks like this (with line breaks inserted for clarity):

```JavaScript
/**
 * @typedef {
 *   Compiler["outputFileSystem"] & {
 *     createReadStream?: import("fs").createReadStream,
 *     statSync?: import("fs").statSync,
 *     lstat?: import("fs").lstat,
 *     readFileSync?: import("fs").readFileSync
 *   }
 * } OutputFileSystem
 */
```

In `@node/types@17.0.14`, `import("fs").statSync` looks like this:

```TypeScript
export const statSync: StatSyncFn;
```

So the `tsc --declaration --emitDeclarationOnly --outDir types` command takes the type of `import("fs").statSync` and emits what it sees as the canonical type, which is `import('fs').StatSyncFn` in Node 17.0.14's typings.

In `@node/types@12.20.51`, `import("fs").statSync` looks like this:

```TypeScript
function statSync(path: PathLike, options?: StatOptions & { bigint?: false | undefined }): Stats;
function statSync(path: PathLike, options: StatOptions & { bigint: true }): BigIntStats;
function statSync(path: PathLike, options?: StatOptions): Stats | BigIntStats;
```

If you add a `devDependency` on `@types/node@12.20.51` to `webpack-dev-server`, the TypeScript compiler now sees the above code block as the type of `import("fs").statSync`. Because there isn't a more canonical type (as there is in newer versions of the Node typings), it simply emits `typeof import("fs").statSync` as the type of `OutputFileSystem.statSync`.

The typings are regenerated upon dependency installation, so having to manually go back and fix them before every publish is begging for the issue to continue coming back. Simply adding a `devDependency` on an older version of `@types/node` fixes this issue permanently and without having to manually tweak compiler output.

alexander-akait: Sounds good, let's move to `devDependencies`

snitin315: Do we add it as devDependency here in webpack-dev-middleware or webpack-dev-server?

iclanton: @snitin315 - here, in the `webpack-dev-middleware` package.

iclanton: Oh looks like I mistyped earlier - yeah I meant `webpack-dev-middleware`, not `webpack-dev-server`

snitin315: @alexander-akait @iclanton Done ✅

alexander-akait (approved):

Issue #1452, opened 2023-02-08
Incorrect content type when using typescript worker in dev server


ericvergnaud:
### Bug report

When creating a typescript worker, the browser fails to load the transpired worker code because its content type is 'video/mp2t' (sourced from mime-db)
See for example: https://stackoverflow.com/questions/74357942/webpack-refuses-execute-script-because-its-mime-type-video-mp2t-is-not-execu
or https://github.com/andywer/threads.js/issues/387

### Actual Behavior

The worker code is correctly transpired, however when returned to the browser, the content type header is set to 'video/mp2t', which causes the browser to fail loading the transpired code.

### Expected Behavior

The content type header should be set to "application/javascript", such that the browser can properly execute it.

### How Do We Reproduce?

Simply create a worker: 
```const worker = new Worker(new URL("./my_worker.ts", import.meta.url));```

**Comment by alexander-akait** [comment]
Can you create reproducible test repo, I think you have a problem with configuration, there is not `ts` content-type

**Comment by ericvergnaud** [comment]
My worker is written in Typescript, so it has the ```ts```extension (not a content-type)
```
new Worker(new URL("./TeknaWorkerThread.ts" , import.meta.url ), { type: "module"});
```
If I use ```.js```in the above example I get the following error:
Module not found: Error: Can't resolve './TeknaWorkerThread.js' in ...

Debugging the dev server, I found that the above call translates into an HTTP GET /static/.../TeknaWorkerThread.SOMEHASH.ts

Then the ts is transpiled to js. But middleware infers the content type from the ts extension.
My proposed fix is precisely to treat as ```js```

I'll try to provide a simple reproducible example in the coming days

ericvergnaud:
Creating a project from scratch with a simple worker does not exhibit the issue.
This could be a side effect of using acejs, needing more exploration.
Closing

alexander-akait:
Just note - `ts` is still part of `js` ecosystem, so after transpiling the worker file should have `js` extensions, there is not `ts` content-type, `ts` is a part of `js`, like you can't have raw react code or vue code for execable files

PR #1454, opened 2023-02-08
Fix issue #1452

ericvergnaud:
… js file, see issue #1452

This PR contains a:

- [X ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

See issue #1452

### Breaking Changes

Not a breaking change

### Additional Info

Root cause could be seen as mime-db not having "ts" as a valid extension for 'application/javascript', however changing that would be incorrect since ts is not js. Rather, we need to cater for the specific scenario where the browser is requesting a ts file but receives a js file as a result of the ts -> js transpilation.

### DIFF
--- a/src/middleware.js
+++ b/src/middleware.js
@@ -140,9 +140,12 @@ function wrapper(context) {
       }
 
       if (!getHeaderFromResponse(res, "Content-Type")) {
-        // content-type name(like application/javascript; charset=utf-8) or false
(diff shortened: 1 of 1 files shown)


ericvergnaud:
Issue isn't reproduced with a simple project. Closing.

PR #1472, opened 2023-03-06
fix: add writeToDisk support for clean: true

Mortaro:
This PR contains a:

- [x ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

I'm the author of [Nullstack](https://nullstack.app/), one of its build modes targets chrome extensions. 
I use the writeToDisk option in development mode to preview the extensions with HMR, but the middleware was never cleaning up the real FS when MemFS unlinked files, causing the output folder to become bloated over time.

### Breaking Changes

There are no breaking changes.

### Additional Info

The solution I added might be a bit hacky, but it does solve this issue that has been open for a minute: https://github.com/webpack/webpack-dev-middleware/issues/861

I have temporarily moved the unstable branch of Nullstack to use [my fork](https://github.com/nullstack/nullstack/commit/2b755b662206053a8aa80f3495e61ac8587d97bb),  but would love to have this fix merged so i can rollback to the official package 💜

### DIFF
--- a/src/utils/setupWriteToDisk.js
+++ b/src/utils/setupWriteToDisk.js
@@ -21,6 +21,20 @@ function setupWriteToDisk(context) {
     (context.compiler).compilers || [context.compiler];
 
   for (const compiler of compilers) {
+    if (compiler.outputFileSystem && compiler.options.output.clean) {
+      /** @type {Compiler["outputFileSystem"]} */
+      const { unlink: originalUnlink } = compiler.outputFileSystem;
+      if (originalUnlink) {
+        compiler.outputFileSystem.unlink = (
+          /** @type {String} */ originalPath,
+          /** @type {(arg0?: null | NodeJS.ErrnoException) => void} */ originalCallback
+        ) => {
+          fs.unlink(originalPath, () => {});
+          originalUnlink(originalPath, originalCallback);
+        };
+      }
+    }
+
     compiler.hooks.emit.tap(
       "DevMiddleware",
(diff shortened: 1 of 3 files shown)


src/utils/setupWriteToDisk.js
```diff
@@ -21,6 +21,20 @@ function setupWriteToDisk(context) {
...
   for (const compiler of compilers) {
+    if (compiler.outputFileSystem && compiler.options.output.clean) {
+      /** @type {Compiler["outputFileSystem"]} */
+      const { unlink: originalUnlink } = compiler.outputFileSystem;
+      if (originalUnlink) {
+        compiler.outputFileSystem.unlink = (
+          /** @type {String} */ originalPath,
+          /** @type {(arg0?: null | NodeJS.ErrnoException) => void} */ originalCallback
+        ) => {
+          fs.unlink(originalPath, () => {});
```
alexander-akait, line 32: Hello, there is racing you should when the first operation finished,  also using `fs.unlink` is not safe, because some developer use custom `outputFileSystem` to avoid file removing, but we remove them and break application, we nee to think deeply how we can handle such case. In theiry we can apply clean plugin before watch and it will with original fs

Mapuppy09:
Add write disk support for clean true

Mapuppy09:
Add clean disk

PR #2238, opened 2026-01-08
chore: remove colorette dependency and update minimum webpack version to 5.101.0

bjohansebas:
**Summary**

With this, we reduce the dependencies we have and use the native helper provided by webpack without a fallback.

chore

Yes, it introduces a breaking change that requires using a webpack version greater than or equal to 101.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -3,7 +3,6 @@
   "language": "en,en-gb",
   "words": [
     "memfs",
-    "colorette",
     "noextension",
     "fullhash",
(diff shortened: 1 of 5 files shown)


alexander-akait:
Thanks, looks good, will be a part of the next major release, we need to prepare dev server for the next major release (and webpack-cli), we are working on it

bjohansebas:
I’ve created the `next` branch, what other changes would you like for the major version?

PR #2259, opened 2026-01-29
feat: add `forwardError` option to enable error forwarding to next middleware

bjohansebas:
**Summary**

While it works fine in Express, I’m not sure how it will behave in other frameworks, so testing is in progress.


src/options.json
```diff
@@ -172,6 +172,11 @@
       "description": "Enable or disable setting `Cache-Control: public, max-age=31536000, immutable` response header for immutable assets (i.e. asset with a hash in file name like `image.a4c12bde.jpg`).",
       "link": "https://github.com/webpack/webpack-dev-middleware#cacheimmutable",
       "type": "boolean"
+    },
+    "forwardError": {
+      "description": "Enable or disable forwarding errors to next middleware.",
+      "link": "https://github.com/webpack/webpack-dev-middleware#forwarderrors",
+      "type": "boolean"
```
alexander-akait: Let's add this to our README too

test/middleware.test.js
```diff
@@ -3610,6 +3609,235 @@ describe.each([
...
+          if (name !== "hapi") {
+            expect(response.text).toBe("error");
+            expect(nextWasCalled).toBe(true);
+          } else {
+            expect(nextWasCalled).toBe(false);
+          }
+        });
+
+        // TODO: why koa and hono don't catch for their error handling when stream emit error?
+        (name === "koa" || name === "hono" ? it.skip : it)(
```
bjohansebas: The behavior here in Hono and Koa is a bit strange, and it has to do with how they handle streams when sending the response. If an error is thrown, it’s not really captured, in fact, Hono would treat it as if `forwardError` were false.

PR #2304, opened 2026-04-05
fix: improve URL handling in getRequestURL function

### LINKED ISSUE #2303: v8 does not respect req.url
### Have you used AI?

No

### Bug Description

In v8 of webpack-dev-middleware, the middleware is trying to serve `req.originalUrl` when it should be trying to serve `req.url`. This is happening because of [this logic](https://github.com/webpack/webpack-dev-middleware/blob/main/src/utils.js#L316-L318). This breaks usage when used with URL rewrites, which is common with middleware like [connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback/tree/main).

### Link to Minimal Reproduction and step to  reproduce

https://github.com/Methuselah96/webpack-dev-middleware-rewrite

Steps:
1. Run `npm install`
2. Run `npm run watch`
3. Navigate to http://localhost:3000/some-path
4. Observe "Cannot GET /some-path" error

### Expected Behavior

The dev middleware should respect the `req.url` and serve `/index.html` instead of returning an error

### Actual Behavior

The dev middleware tries to load `/some-path` and fails

### Environment

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i5-13500
    Memory: 7.01 GB / 31.65 GB
  Binaries:
    Node: 24.11.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.9.2 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.6.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.15.9 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Chrome: 146.0.7680.165
    Edge: Chromium (140.0.3485.54)
    Firefox: 149.0 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
  Packages:
    html-webpack-plugin: ^5.6.6 => 5.6.6 
    webpack: ^5.105.4 => 5.105.4 
    webpack-cli: ^7.0.2 => 7.0.2 
    webpack-dev-middleware: ^8.0.2 => 8.0.2
```

### Is this a regression?

Yes (please specify version below)

### Last Working Version

v7.4.5

### Additional Context

_No response_

bjohansebas:
**Summary**

fixed #2303 

**Use of AI**

yes!

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -312,9 +312,27 @@ function getRequestURL(req) {
   if (typeof req.getURL === "function") {
     return req.getURL();
   }
-  // Fastify decodes URI by default, Our logic is based on encoded URI
-  else if (typeof req.originalUrl !== "undefined") {
-    return req.originalUrl;
+
+  // Fastify decodes URI by default, our logic is based on encoded URI.
+  // req.originalUrl preserves the original encoded URL; req.url may be
+  // modified by middleware (e.g. connect-history-api-fallback), in which
+  // case we use req.url instead.
+  if (typeof req.originalUrl !== "undefined") {
+    // If req.url is just the decoded form of req.originalUrl (Fastify behavior),
+    // return the original encoded URL. Otherwise middleware modified req.url.
+    try {
+      if (
+        req.url === req.originalUrl ||
+        (typeof req.url !== "undefined" &&
+          decodeURI(req.originalUrl) === req.url)
+      ) {
+        return req.originalUrl;
+      }
+    } catch {
+      // decodeURI can throw on malformed sequences, fall through
+    }
+
+    return req.url;
   }
 
   return req.url;
--- a/src/utils.js
+++ b/src/utils.js
@@ -255,6 +255,7 @@ function parseTokenList(str) {
  * @property {(() => string | undefined)=} getMethod get method extra method
  * @property {(() => string | undefined)=} getURL get URL extra method
  * @property {string=} originalUrl an extra option for `fastify` (and `@fastify/express`) to get original URL
+ * @property {string=} id an extra option for `fastify` (and `@fastify/express`) to get ID of request
  */
 
 /**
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -312,9 +312,27 @@ function getRequestURL(req) {
...
+      if (
+        req.url === req.originalUrl ||
+        (typeof req.url !== "undefined" &&
+          decodeURI(req.originalUrl) === req.url)
+      ) {
+        return req.originalUrl;
+      }
+    } catch {
+      // decodeURI can throw on malformed sequences, fall through
+    }
```
alexander-akait, line 326: It this problem only for fastlify? `decodeURI` is not fast, so it is bad for performance

bjohansebas: Yep, it's unique to Fastify, or at least I don't remember Express having the same issue. Fastify uses its own logic instead of using decodeURI https://github.com/fastify/middie/pull/245

alexander-akait: in this case let's add a pseudo API for fastify here and implement this logic only for fastify, don't want to lose perfomance for other frameworks

bjohansebas: I’m not sure what steps to follow. It would be great if Fastify exposed something like `req.raw` on the request (since it currently doesn’t, which is a bit odd—but I’m not a Fastify expert). That way we could quickly detect it, since neither Express nor Connect have that property. However, Hono also has a `getURL` method, and since our logic checks `getURL`, it would never reach the fallback logic using `req.raw`.

Now the real issue is that Fastify *does* decode `req.url`, while Express and others don’t.

We could add logic like this:

```js
if (typeof req.originalUrl !== "undefined") {
  if (req.url === req.originalUrl) {
    return req.originalUrl;
  }

  // Some frameworks can expose decoded req.url while keeping
  // encoded req.originalUrl.
  if (typeof req.url !== "undefined" && req.originalUrl.includes("%")) {
    try {
      if (decodeURI(req.originalUrl) === req.url) {
        return req.originalUrl;
      }
    } catch {
      // Ignore malformed URI sequences and fall back to req.url.
    }
  }
}
```

Or we could create a wrapper for Fastify, but I’m not sure how reliable it would be to correctly detect that Fastify is running underneath.

Another option would be to open an issue in `@fastify/express` to have them expose `raw`, so we can better detect Fastify internally.

What would you recommend without hurting performance too much?

I’m thinking about other ways to detect it—I’ll see what I come up with by tomorrow night

alexander-akait: I see the problem... Made refactor to use `req.id` to apply this logic only for fastify

alexander-akait: > Or we could create a wrapper for Fastify, but I’m not sure how reliable it would be to correctly detect that Fastify is running underneath.

I think in future we can do it and implement built-in fastify support without the fastify-express layer, but to be honestly fastify is not very popular, so I think we will use fastify-express until it was supported, when it will be deprecated we will implement own wrapper

> Another option would be to open an issue in @fastify/express to have them expose raw, so we can better detect Fastify internally.

It will be fine, let's try it, if it will be fixed we will revert this code and will use `raw`, so feel free to open an issue

alexander-akait (approved):

src/utils.js
bjohansebas: I didn’t know `req.id` existed, that’s new to me.

PR #2321, opened 2026-05-16
feat: implement hot module replacement middleware

bjohansebas:
**Summary**

**Use of AI**


src/hot.js
```diff
@@ -0,0 +1,356 @@
...
+function publishStats(action, statsResult, eventStream, log, statsOptions) {
+  const resultStatsOptions = {
+    all: false,
+    cached: true,
+    children: true,
+    modules: true,
+    timings: true,
+    hash: true,
+    errors: true,
+    warnings: true,
```
alexander-akait, line 223: Ideally we need only require stats, it is slow down build

src/hot.js
```diff
@@ -0,0 +1,356 @@
...
+    ...(statsOptions && typeof statsOptions === "object" ? statsOptions : {}),
+  };
+
+  /** @type {EXPECTED_ANY[]} */
```
alexander-akait: Let's avoid EXPECTED_ANY, we have types here

src/hot.js
bjohansebas: what do you mean?

PR #2323, opened 2026-05-16
feat: add browser client runtime for HMR

bjohansebas:
**Summary**

**Use of AI**

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -5,10 +5,12 @@ on:
     branches:
       - main
       - next
+      - hot-middleware
   pull_request:
     branches:
       - main
       - next
+      - hot-middleware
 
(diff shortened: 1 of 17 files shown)


alexander-akait:
@bjohansebas Please don't merge such things in future without approving, at least 1 approve, I don't make it strict to be more flexibility, but it doesn't mean we should ignore it, I don't merge then because we need to change to architecture problems on webpack and dev server side

bjohansebas:
this isn't merged into `main`, it's only in another branch (#2322), so any changes can still be made there.

alexander-akait:
Yeah, I see it, just for future, with other branches will be good to make the right architecture too, but it was already merged, so we will work with branch together, I don't like it because it makes a big diff and often unreadable

PR #2348, opened 2026-07-14
fix(client): avoid double slash when joining dynamicPublicPath

bjohansebas:
When output.publicPath ends with a slash (the common case) and the SSE path starts with one, the naive concatenation produced URLs like `https://host//__webpack_hmr`, which never match the server-side pathname check.

Ref webpack/webpack-hot-middleware#366

**Summary**

**Use of AI**


client-src/index.js
```diff
@@ -57,7 +57,13 @@ function setOverrides(overrides) {
...
 
   if (overrides.dynamicPublicPath) {
-    options.path = __webpack_public_path__ + options.path;
+    // Avoid a double slash when the public path has a trailing slash and the
+    // SSE path a leading one (e.g. "https://host/" + "/__webpack_hmr").
+    const publicPath = __webpack_public_path__;
+    options.path =
+      publicPath.endsWith("/") && options.path.startsWith("/")
+        ? publicPath + options.path.slice(1)
+        : publicPath + options.path;
```
alexander-akait: Hm, we should be parity with webpack logic at runtime, if we connect two slashes it means we should do it here and maybe handle it on options level or documentation, in some rare cases two slashes can be used (nginx can rewrite requires and use two slashes, but I am sure no one really use it)

alexander-akait (approved):

PR #2353, opened 2026-07-14
feat(client): bring the error overlay to parity with webpack-dev-server

bjohansebas:
- Render inside an about:blank iframe and style exclusively through the CSSOM so the overlay works under a strict style-src CSP; inline styles from ansi-html are re-applied via style.cssText.
- Support Trusted Types: innerHTML writes go through a policy (configurable via overlayTrustedTypesPolicyName).
- Capture uncaught runtime errors and unhandled rejections in the overlay (overlayRuntimeErrors, default true), with the same React error boundary heuristic as dev-server.
- Inline the HTML entity encoder and drop the html-entities dependency.
- Expose the overlay as a standalone subpath export (webpack-dev-middleware/client/overlay) so webpack-dev-server can reuse it.

Ref webpack/webpack-hot-middleware#457
Ref webpack/webpack-hot-middleware#184

**Summary**

**Use of AI**


alexander-akait:
Let's postpone it right now, ideally firstly we should merge hot middleware and then merge overlay and clients from (in dev server we have web socket, in hot middleware we use SSE, in webpacj for lazy compilation we use SSE), so we need to think how we can union all these stuff and provide an option for developers to set to use SSE or websocket, feel free to experiments

alexander-akait (approved):
Looks good, as a first step after the release we should start to use overlay from there in dev server

bjohansebas:
Yep, once it's on npm, we can migrate everything, so I'm going to merge this. 🙂

The overlay already works with both WebSockets and SSE, so there aren't any issues there. The only thing I did afterward was fix a problem with the multi-compiler: in my latest commits for the sse, it should now display errors from both compilers instead of only the last one.

PR #2358, opened 2026-07-21
feat(hot): add a building indicator with optional compilation progress

bjohansebas:
The client shows a small badge (shadow DOM, CSSOM-only styles) while a rebuild is in progress, enabled by default (disable with `?progress=false` on the client entry). When the server enables the new `hot.progress` option, webpack's ProgressPlugin publishes throttled `progress` events over SSE (deduplicated by rounded percent, reset on each new build) and the badge renders a CSP-safe SVG progress ring with the compilation percentage.

The palette is shared with the error overlay through a common theme module, and the indicator is exposed as a standalone subpath export (webpack-dev-middleware/client/indicator) so other tooling can reuse it. Both the indicator and the overlay now bail out safely when `document.body` does not exist yet.

Ref webpack/webpack-hot-middleware#167

**Summary**

**Use of AI**

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,13 @@ Default: `10000`
 
 Heartbeat interval (in milliseconds) used to keep the SSE connection alive when no compilation events are produced.
 
+#### `hot.progress`
+
+Type: `Boolean`
+Default: `undefined`
+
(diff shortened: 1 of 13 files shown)


alexander-akait (approved):
The same I have feeling that dev server overlay support it, anyway feel free to merge

bjohansebas:
> The same I have feeling that dev server overlay support it, anyway feel free to merge

Yeah, but we only export it so both `dev-server` and anywhere else can use the same component. That way we maintain a single shared component instead of having to maintain multiple copies

PR #2359, opened 2026-07-21
feat(client): paginate the overlay problems

bjohansebas:
The overlay now shows one problem at a time, with a header row holding the problem badge and prev/next navigation (colored after the problem type), a page counter and arrow-key navigation. Each new problem set starts at its first page, accumulated runtime errors land on the newest one, and identical re-published sets (e.g. another bundle of a multi-compiler syncing) keep the current page. Enabled by default; disable with the overlay option's `paginate` extension (`?overlay={"paginate":false}`) to render the full list.

Also ignore click targets re-rendered away mid-dispatch in the backdrop-dismiss listener, which otherwise closed the overlay when clicking the navigation buttons.

imagen
imagen

**Summary**

**Use of AI**


alexander-akait (approved):
I have a feeling that we already support in dev server overlay, anyway we are doing all step by step, so feel free to merge

PR #2360, opened 2026-07-21
fix(client): reload when an accept handler errors during apply

bjohansebas:
An error thrown inside a module.hot.accept handler was logged and ignored, leaving the page running stale code with no recovery. When the `reload` option is enabled (the default), onErrored now falls back to a full page reload, like the other unrecoverable update paths.

The HMR runtime and page-reload calls are isolated behind small utils so process-update finally has direct test coverage. A missing HMR runtime is now reported with a single actionable error instead of throwing at bundle evaluation.

Ref webpack/webpack-hot-middleware#334

**Summary**

**Use of AI**

### DIFF
--- a/client-src/process-update.js
+++ b/client-src/process-update.js
@@ -90,6 +90,7 @@ export default function applyUpdate(hash, options) {
    * @param {Error} err error
    */
   function handleError(err) {
+    // @ts-expect-error function declarations are hoisted, so the `!hot` guard narrowing is lost
     if (hot.status() in failureStatuses) {
       log.warn("Cannot check for update (Full reload needed)");
       log.warn(err.stack || err.message);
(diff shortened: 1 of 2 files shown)


client-src/process-update.js
```diff
@@ -52,15 +57,33 @@ function upToDate(hash) {
...
+  if (!maybeHot) {
+    // Logged (once) instead of thrown: this runs inside the SSE message
+    // handler, whose catch would mislabel a throw as an invalid message.
+    if (!loggedRuntimeMissing) {
+      loggedRuntimeMissing = true;
+      log.error(
+        "[HMR] Hot Module Replacement is disabled. " +
+          "Add HotModuleReplacementPlugin to the webpack configuration.",
+      );
+    }
```
bjohansebas, line 71: I think it's better for the developer experience to do it this way, but let me know if you think otherwise

client-src/process-update.js
```diff
@@ -52,15 +57,33 @@ function upToDate(hash) {
...
+    return;
+  }
+
+  const hot = maybeHot;
```
bjohansebas: I don't really like this, but it's needed to satisfy TypeScript

alexander-akait: Feel free to use @ ts-expected-error to avoid extra runtime code

alexander-akait (commented):
A small comment

PR #2412, opened 2026-09-18
ci: cancel superseded pull request runs

alexander-akait:
`nodejs.yml` already cancelled superseded jobs, but through three per-job concurrency groups that also cancel pushes to `main`, `next` and `hot-middleware`. This replaces them with one workflow-level group that only cancels pull request runs.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

## Why the branch pushes should not be cancelled

That is the actual bug here. `main`, `next` and `hot-middleware` are built on push, those runs publish coverage, and two merges in quick succession cancel the first one's. A cancelled run also hides a breakage that is already on the branch, which is exactly when you want to know.

## Why one workflow-level group rather than fixing the three in place

Adding the condition to the existing blocks would have been a smaller diff, but the per-job arrangement has two other costs:

- **It cancels job by job.** Each old job dies only as its replacement is queued; a workflow-level group cancels the whole superseded run at once. This workflow is 14 jobs (12 test cells across three operating systems and four Node.js versions, plus `lint` and `client`).
- **Every matrix dimension has to be repeated in the group key.** `test` currently spells out `os`, `node-version` and `webpack-version`; a dimension added later and forgotten would make cells cancel each other, which is a quiet and confusing failure. Keying on the workflow and ref cannot drift.

Happy to switch to the minimal version (keep the per-job groups, just add the condition) if you would rather keep the existing shape.

## Verification

`actionlint` 1.7.7 is clean on all four workflows, and I checked that this means something by typo'ing the expression to `github.event_nam`, which it catches at exactly that line. The file is also Prettier- and cspell-clean using this repo's own configs.

`release.yml` is untouched: it already uses `concurrency: ${{ github.workflow }}-${{ github.ref }}`, whose string form leaves `cancel-in-progress` false, so releases queue instead of cancelling. `dependabot.yml` and `dependency-review.yml` are untouched too — short jobs, and cancelling the Dependabot one halfway could leave a pull request without the auto-merge it was about to enable.

The same change is going to `enhanced-resolve` (webpack/enhanced-resolve#673) and `webpack-dev-server` (webpack/webpack-dev-server#5742).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF

---
_Generated by [Claude Code](https://claude.ai/code/session_016aGHrb1YaEvaGwNELGEHjF)_

## Summary by CodeRabbit

* **Chores**
  * Improved automated workflow run management by grouping runs according to the workflow and branch or reference.
  * Pull request workflow runs now cancel superseded in-progress runs to reduce redundant checks.
  * Other workflow runs, including builds on key branches, continue independently without automatic cancellation.
  * This provides more predictable validation results while avoiding unnecessary duplicate processing.


alexander-akait:
Pushed a correction in 9acaad2 before anyone spent time reviewing the first version — the review on the sibling PR (webpack/enhanced-resolve#673) caught a real flaw that this PR had too.

`cancel-in-progress: false` protects only the run that is already executing. The concurrency group still holds at most one *pending* run, and GitHub evicts that pending run whenever a newer one enters the group, whatever the flag says — *"any existing pending job or workflow in the same concurrency group will be canceled and the new queued job or workflow will take its place."*

So keying every run on the ref did not actually deliver what this PR claims. With one `main` run executing and a second waiting, a third push would have evicted the second — exactly the "two merges in quick succession" case the change exists to protect, and it would have looked correct while failing.

The group is now:

```yaml
group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref || github.run_id }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Pull requests still group by ref and supersede as described above; every other run gets a group of one and can be neither cancelled nor evicted. `actionlint` 1.7.7, Prettier and cspell are clean on the result using this repo's own configs.

---
_Generated by [Claude Code](https://claude.ai/code)_

